PR #11, opened 2016-06-15
Fix deadline counters

marlabrizel:
Deadline counters were not appropriately padding single-digit entries, which was awkward.

Single-digit hours and minutes are now left-padded. Double-digits are unaffected.


jayzes:
:metal:

PR #24, opened 2016-07-20
Submitters can edit their session proposals

### LINKED ISSUE #17: Submitters should be able to propose updates to their session title/description

marlabrizel:
Limited to title, description, and presenter notes only. User edits kicked over to ActiveAdmin for approval and must be manually accepted before going live on the site. An email is triggered to track chairs whenever someone makes a suggested update.

Closes #17

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -88,6 +88,7 @@ class SubmissionsController < ApplicationController
     @submission.update(proposed_updates: submission_params)
     if @submission.save
       flash[:notice] = 'Thanks! Your changes have been submitted and are pending review.'
+      @submission.notify_track_chairs_of_update
       redirect_to mine_submissions_path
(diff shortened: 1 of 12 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -80,6 +81,19 @@ def create
     end
   end
 
+  def edit
+  end
+
+  def update
+    @submission.update(submission_params)
```
jayzes: I think we should figure out a way to put these updates in a holding pen to allow for approval before they go live. Maybe we could add a new JSON column that stores pending changes, then add an admin action to view/approve them?

marlabrizel: Yeah agree - this is sort of a naive implementation that would provide a backdoor for people to self-grant submission extensions (among other potentially undesirable scenarios)

PR #38, opened 2016-08-16
Fix HelpScout nil errors

marlabrizel:
Add guard clause so that helpscout doesn't blow up trying to pull submissions or registration for a non-existent user


jayzes:
:shipit:

PR #61, opened 2016-09-13
Add an iCal location to submission so we can have the address in the iCal schedule

tghw:
That way, if you import into Google Maps, you can get more reliable directions.


jayzes:
Awesome - thanks @tghw!

PR #105, opened 2017-07-18
Convert homepage CTAs to be dynamic

jayzes:
- Bring in `liquid` to interpolate dynamic variables (though they get cached so we’ll need to be aware of dates)
- Convert a few more feature flags over to `EventSchedule`
- Add an admin UI
- Seed existing CTAs


jayzes:
Going to self-review and merge but @marlabrizel if you want to :eyes: when you have a chance that'd be awesome.

Issue #122, opened 2017-08-13
Create DSW Youth Pass Landing Page


ryefinegan:
**Landing Page Copy:**

Young Entrepreneurs Wanted!

For the first time, Junior Achievement and Young Americans Center for Financial Education are partnering with Denver Startup Week to provide exciting opportunities for Denver’s young entrepreneurs! 

YouthBiz Marketplace and Expo - Tuesday, September 26th, 3 - 4:30pm at Basecamp
Do you want to take your business plan or company to the next level?  Come show off your stuff at Denver Startup Week's Basecamp! The YouthBiz Marketplace and Expo is the perfect place for young entrepreneurs to come pitch their ideas and showcase their companies. Ages 6-18.

APPLY NOW - https://docs.google.com/forms/d/1L-3kmDo6vX5Bttbpzg0xjwKetmD8Sw8iSvDoKiOYWrM/edit?ts=598df8bb 

Youth Early Access - Thursday, September 28th, 9am - 3pm
Participate in a first-of-its-kind Denver Startup Week field trip experience for high school students!  Hear from some of Denver's most successful entrepreneurs before taking off on a customized Denver Startup Week tour to see an array of awesome DSW sessions around town. Adult chaperones will accompany youth to all sessions.

Click here to register a group of students - https://fs22.formsite.com/JuniorAchievementRMI/form30/index.html
Click here to register an individual student - https://fs22.formsite.com/JuniorAchievementRMI/form31/index.html

PR #131, opened 2017-08-18
Add youth landing page

jayzes:
Fixes https://github.com/denverstartupweek/dsw-site/issues/122


jayzes:
Just needs an updated marketplace form link and it should be good to go live.

PR #143, opened 2017-09-05
Populate event-specific sponsors

jayzes:


jayzes:
Going to merge this and proceed with moving these logos to S3

PR #215, opened 2017-10-29
Send email when session updates are accepted

### LINKED ISSUE #162: Accepting proposed changes to a sessions should trigger a notification back to the submitter
That way they know the changes have taken effect.

marlabrizel:
Sends an email notification to all session submitters for a given session when proposed session update are accepted by a DSW admin.

Closes #162

### DIFF
--- a/app/mailers/notifications_mailer.rb
+++ b/app/mailers/notifications_mailer.rb
@@ -18,6 +18,13 @@ class NotificationsMailer < ApplicationMailer
     mail to: chair.email, subject: "#{@submission.submitter.name} has proposed an update needing your approval for the #{@track.name} track."
   end
 
+  def notify_of_update_acceptance(submission)
+    @submission = submission
+    mail to: [ @submission.contact_emails,
+               @submission.submitter.email ].flatten.uniq,
+         subject: 'Your proposed session updates have been accepted'
+  end
+
   def confirm_new_submission(submission)
     @submission = submission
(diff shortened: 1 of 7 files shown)


app/models/submission.rb
```diff
@@ -294,48 +294,51 @@ def send_venue_match_email!
...
     sent_notifications.create! kind: SentNotification::THANKS_KIND,
-                               recipient_email: message.to.join(', '),
-                               body: message.message.to_yaml
+                               recipient_email: message.to.join(', ')
   end
 
   def promote_updates!
     with_lock do
       update((proposed_updates || {}).merge(proposed_updates: nil))
+      notify_submitter_of_update_acceptance!
```
jayzes: It might be good to move this outside of the `with_lock` block - feels like we wouldn't want this to roll back if the e-mail send failed

app/mailers/notifications_mailer.rb
```diff
@@ -18,6 +18,13 @@ def notify_of_submission_update(chair, submission)
     mail to: chair.email, subject: "#{@submission.submitter.name} has proposed an update needing your approval for the #{@track.name} track."
   end
 
+  def notify_of_update_acceptance(submission)
+    @submission = submission
+    mail to: [ @submission.contact_emails,
+               @submission.submitter.email ].flatten.uniq,
```
jayzes: Wonder if this pattern is frequent enough that we should move it to a helper?

Issue #220, opened 2017-10-30
Schedule Page


ryefinegan:
Can we change the top title to "SCHEDULE 2017 - SEPTEMBER 25 - 29" and put the below text at the top of the schedule page?

"Thank you for an incredible Denver Startup Week 2017!  Information about submitting sessions for Denver Startup Week 2018 will be shared early next year."

PR #221, opened 2017-12-05
 Make yearly schedules admin-editable

jayzes:
Make yearly schedules admin-editable

 - Use a new model with a JSONB column to easily add dates
- Replace EventSchedule with AnnualSchedule
- Seed 2017 dates
- Display next year’s dates on the homepage if we’re after the week and they’re set
- Set up an admin

This is the first part of a fix for https://github.com/denverstartupweek/dsw-site/issues/220.


app/models/annual_schedule.rb
```diff
@@ -0,0 +1,134 @@
+class AnnualSchedule < ApplicationRecord
+
+  CFP_CYCLE = 'cfp'.freeze
```
marlabrizel: let's add `# frozen_string_literal: true` to the top of this file so that we don't have to call `.freeze` on each constant

app/views/site/index.html.slim
```diff
@@ -1,7 +1,8 @@
 section.hero
-  - unless EventSchedule.in_week?
-    / h1.countdown #{time_remaining_to_deadline(EventSchedule::WEEK_START_DATE)}
-  h2 September 25&#8211;29, 2017
+  - if AnnualSchedule.post_week? && AnnualSchedule.next.present?
+    h2 #{AnnualSchedule.next.week_start_at.strftime('%B %-d')}&#8211;#{AnnualSchedule.next.week_end_at.strftime('%-d, %Y')}
```
marlabrizel: is it possible to change the dash encoding to its more readable equivalent?

app/views/site/index.html.slim
```diff
@@ -1,7 +1,8 @@
...
+  - else
+    h2 #{AnnualSchedule.current.week_start_at.strftime('%B %-d')}&#8211;#{AnnualSchedule.current.week_end_at.strftime('%-d, %Y')}
```
marlabrizel: same here re: unicode

app/models/annual_schedule.rb
jayzes: Done

app/views/site/index.html.slim
jayzes: Yup, done!

app/views/site/index.html.slim
jayzes: Also done.

PR #242, opened 2018-03-13
Migrate FAQ to Helpscout docs

### LINKED ISSUE #232: Migrate FAQ to be managed via Helpscout's knowledge base
We can move the existing questions/answers over to the current internal Docs instance and then access them via the API to render that page: https://developer.helpscout.com/docs-api/. Will probably have to do a bit of caching so we don't hit limits and/or to make that page load performant.

This would have the advantage of making it easier to manage that content (@ryefinegan or anyone else with Helpscout access can do it) and making it easier to access for folks doing support.

jayzes:
Fixes #232


jayzes:
@marlabrizel this is ready for initial review, I think. I did have to rebase off master to get Sidecloq integration in, so FYI.

lib/helpscout/client.rb
```diff
@@ -0,0 +1,27 @@
+module Helpscout
+  class Client
+    include HTTParty
+
+    BASE_URI = 'https://docsapi.helpscout.net/v1'.freeze
+
+    base_uri BASE_URI
+
+    def articles
+      data = self.class.get('/categories/5aa2b5f02c7d3a75495181d5/articles', basic_auth: auth)
```
marlabrizel: we should decide how we want to extract this category ID in lieu of hardcoding

marlabrizel:
Do we want to deal with category id abstraction now or is it more important to get this out ahead of the kickoff and revisit that bit later @jayzes? if latter, I will create an issue to handle. otherwise, lgtm.

jayzes:
I'm fine with getting it this out the door now and punting the category ID for a day or two. Would be good to do that as part of getting the panel picker FAQ (https://www.denverstartupweek.org/panel-picker/faq) moved over.

PR #245, opened 2018-03-18
Adds Code of Conduct

### LINKED ISSUE #236: Add Code of Conduct
@marlabrizel has a template to use, I think.

Some interesting stuff here too: https://wearemovingforward.github.io/

Code of conduct needs to live in the following places:
- Site footer partial - accessible from every page
- Mandatory "I have read and agree" checkbox on CFP submission page with link out to COC
- Main nav

marlabrizel:
Adds Code of Conduct to site
Adds permalink to footer nav
Requires CoC acceptance during CFP and Registration flows

Does some unrelated cleanup around removing unused views

closes #236


app/assets/stylesheets/redesign/views/code_of_conduct.sass
```diff
@@ -0,0 +1,19 @@
+@import ../config
+
+section.common
+  .section-content
```
jayzes: I think we need to fully scope this within that action-specific selector so it doesn't clobber other styles. I'll take a pass.

app/controllers/submissions_controller.rb
```diff
@@ -63,7 +63,7 @@ def search
   end
 
   def new
-    @submission = Submission.new(contact_email: current_user.try(:email))
+    @submission = Submission.new(contact_email: current_user.try(:email), coc_acknowledgement: false)
```
jayzes: I don't think this is needed, unless I'm missing something? I think this attribute is `nil` by default and won't validate.

jayzes:
@marlabrizel this should be good to merge now

PR #252, opened 2018-03-23
Allow clusters to be toggled active/inactive

### LINKED ISSUE #251: Clusters should be active/inactive
- Allow program clusters to be marked as active/inactive. New clusters default to active if not otherwise specified
- Remove cluster choice from CFP form
- Active Admin updates:
  - Admin can assign a cluster to a session (this should be unchanged from previous functionality)
  - Admin can only choose from active clusters; inactive clusters are not visible as an assignment choice
   - Admin can deactivate/activate clusters from within Active Admin

marlabrizel:
Removes cluster selection as an option on the CFP. Also opportunistically removes question about URM status.

Adds field to clusters wherein they can be toggled active/inactive. Administerable through admin panel. Only currently active clusters can be selected & added to a session.

Closes #251

### DIFF
--- a/app/admin/clusters.rb
+++ b/app/admin/clusters.rb
@@ -2,7 +2,7 @@ ActiveAdmin.register Cluster do
 
   menu parent: 'Sessions', priority: 1
 
-  permit_params :name, :description
+  permit_params :name, :description, :is_active
 
   index do
     selectable_column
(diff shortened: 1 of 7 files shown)


marlabrizel:
@jayzes depends on how we want to interact with clusters. If we scope it to only showing actives, then either you or I will need to go into the db to re-toggle anything we want to bring back. Thoughts?

PR #257, opened 2018-04-01
issue- 253 "Selection" page should link to previous years' schedules

GabrielAfflitto:
I added a link to the previous years schedules from the selection page, this functionality has been tested as well


jayzes:
I think the build is failing due to the removal of `package.json` - looks like Webpacker needs that to configure itself properly.

GabrielAfflitto:
It's giving me an undefined for enviroment.plugins.set in environment.js

GabrielAfflitto:
There's also a dump.rb file that I haven't committed. But I think that has to do with redis

jayzes:
Ah, I see. The README was a bit unclear on setup instructions for the webpacker/yarn side of things - I just pushed up some clarification there. Sorry for the back and forth.

A few things to do to get your local setup in a good place and this PR ready to merge:

- Install yarn globally with `brew install yarn`
- Reset `package-lock.json`, `package.json`, `yarn.lock`, and `spec/support/env.rb` to what they are on `master`. I think `git checkout master -- <path>` for each should do the trick. The local changes for each aren't needed and I believe are actually causing some of the Webpacker build.
- Run `bin/yarn install` to install JS dependencies with the new global Yarn and `package.json` from master
- Remove `dump.rdb` (it's a local database dump and shouldn't be in the repo. I've added it to the .gitignore in master](https://github.com/denverstartupweek/dsw-site/blob/master/.gitignore#L13))

If you don't mind giving those steps a try and letting me know how it goes, that'd be awesome!

jayzes:
👍

GabrielAfflitto:
@jayzes Hey we noticed that you wanted a link to the previous years schedule as well, but we were not sure where to create the link or what the link would be? Let me know and we can add that as well so this issue can be fully completed

jayzes:
Ah, yup, that would be on this page (https://www.denverstartupweek.org/program/previous) - the template should be in `site/program/previous`. I didn't specifically have copy in mind for it, though maybe we could reuse some existing styles and just create a link with an arrow that directs users to the previous year's schedule("EXPLORE FULL SCHEDULE")? Something like what we have under each paragraph in blue on https://www.denverstartupweek.org/program.

PR #265, opened 2018-04-06
update contact info

marlabrizel:
this is mostly a hotfix until we can get all areas of the faq served by helpscout


jayzes:
We should probably eventually make this dynamic too.

Issue #258, opened 2018-04-02
Dates on selection page should be dynamic (from schedule)


jayzes:
Right now they're static content: https://www.denverstartupweek.org/program/selection

PR #266, opened 2018-04-10
Dynamic dates, issue #258

IlinDmitry:
Hello.
I'm practicing Rails. My first PR in this project.

### DIFF
--- a/app/views/site/program.html.slim
+++ b/app/views/site/program.html.slim
@@ -43,10 +43,6 @@ section.common
 section.common
   .section-content.sm-centered
     h1 Key Dates for #{AnnualSchedule.current.year}
-    h6 Kickoff Event: March 21st, 2018
-    h6 Call for Proposals: April 4th - 30th, 2018
-    h6 Voting on Sessions: May 2nd to 18th,  2018
-    h6 Events Announced and Registration Open: July 2018
-    h6 Denver Startup Week: September 24th – 28th
+    = render partial: 'layouts/shared/schedule'
 
 = render partial: 'layouts/shared/page_footer'
--- a/app/views/site/program/selection.html.slim
+++ b/app/views/site/program/selection.html.slim
@@ -8,12 +8,8 @@ section.common.sm-centered
 
 section.common.sm-centered
   .section-content
-    h1 Key Dates for 2018
-    h6 2018 Kickoff Event: March 21st, 2018
-    h6 Call for Proposals: April 4th - 30th, 2018
-    h6 Voting on Sessions: May 2nd to 18th,  2018
-    h6 Events Announced and Registration Open: July 2018
-    h6 Denver Startup Week: September 24th – 28th
+    h1 Key Dates for #{AnnualSchedule.current.year}
+    = render partial: 'layouts/shared/schedule'
 
 section.common.sm-centered
   .section-content
--- a/app/views/layouts/shared/_schedule.html.slim
+++ b/app/views/layouts/shared/_schedule.html.slim
@@ -1,0 +1,14 @@
+ ul
+    li
+      h6 #{AnnualSchedule.current.cfp_open_at.to_s(:month_day)} - Panel Picker opens for submissions
+    li
+      h6 #{AnnualSchedule.current.cfp_open_at.to_s(:month_day)} - Call for submissions closes
+    li
+      h6 #{AnnualSchedule.current.voting_open_at.to_s(:month_day)} - Public voting/feedback on submissions begins
+    li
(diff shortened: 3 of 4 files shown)


app/controllers/site_controller.rb
```diff
@@ -3,6 +3,50 @@ class SiteController < ApplicationController
   respond_to :html
 
   def index
+    a_shed = AnnualSchedule.current
+    @shedule = []
```
marlabrizel: Can you please correct the spelling here? `@shedule` -> `@schedule`

app/controllers/site_controller.rb
```diff
@@ -3,6 +3,50 @@ class SiteController < ApplicationController
   respond_to :html
 
   def index
+    a_shed = AnnualSchedule.current
```
marlabrizel: Hi! We like to use full-name variables here to improve readability. Would you mind renaming this variable to something more descriptive? I would suggest `current_schedule`.

app/controllers/site_controller.rb
```diff
@@ -3,6 +3,50 @@ class SiteController < ApplicationController
...
+    @shedule = []
+
+    unless a_shed.nil?
```
marlabrizel: This is probably too much logic to have in the controller. Can you please push this down into the appropriate model? You will probably want to break it up into multiple, composable methods. You may also wish to consider refactoring this logic away from the heavy use of conditionals.

app/controllers/site_controller.rb
```diff
@@ -34,4 +78,14 @@ def page_partial
     template_name = "site/#{params[:page].presence || 'index'}".underscore
     lookup_context.find(template_name).virtual_path
   end
+
+  def start_date(date_start)
```
marlabrizel: These methods also probably don't belong in the controller.

app/views/site/program.html.slim
```diff
@@ -44,9 +44,7 @@ section.common
   .section-content.sm-centered
     h1 Key Dates for #{AnnualSchedule.current.year}
     h6 Kickoff Event: March 21st, 2018
-    h6 Call for Proposals: April 4th - 30th, 2018
-    h6 Voting on Sessions: May 2nd to 18th,  2018
-    h6 Events Announced and Registration Open: July 2018
-    h6 Denver Startup Week: September 24th – 28th
+    - @shedule.each do |sh|
```
marlabrizel: can you please fix the spelling and variable naming here? This should probably be something like `@current_schedule.each do |milestone|`

app/views/site/program/selection.html.slim
```diff
@@ -8,12 +8,10 @@ section.common.sm-centered
...
 section.common.sm-centered
   .section-content
-    h1 Key Dates for 2018
+    h1 Key Dates for #{AnnualSchedule.current.year}
     h6 2018 Kickoff Event: March 21st, 2018
-    h6 Call for Proposals: April 4th - 30th, 2018
-    h6 Voting on Sessions: May 2nd to 18th,  2018
-    h6 Events Announced and Registration Open: July 2018
-    h6 Denver Startup Week: September 24th – 28th
+    - @shedule.each do |sh|
```
marlabrizel: can you please fix the spelling and variable naming here? This should probably be something like @current_schedule.each do |milestone|

jayzes:
Hi @IlinDmitry, thanks for the PR!

I think we may be able to accomplish this with a bit less code using a pattern like the one in this view: https://github.com/denverstartupweek/dsw-site/blob/fe99ebff9b3907db7ff55f92c95b66a13ee18bd4/app/views/layouts/shared/faq/_timeline.html.slim#L10

That way we don't have to assemble the date string on every request to a static page (even ones that don't necessarily render that data) and we can reuse the existing helpers and formats.

IlinDmitry:
Hello!
Thank you all for review.
@jayzes, you mean like this? Should i also add tests for it? Because  i didn't see any tests for `/home/dima/RubymineProjects/dsw-site/app/views/site/get_involved/faq.html.slim`

jayzes:
@IlinDmitry this looks great! I like that you pulled that repeated content into a partial.

On the test front, we've built a few smoke tests that help provide some quick verification that these content-oriented pages are syntactically correct and render properly. You can see those at https://github.com/denverstartupweek/dsw-site/blob/master/spec/features/content_pages_spec.rb. They're not amazingly great tests, but they do help catch egregious errors, so there's some benefit. It looks like the program page is already covered by that existing suite (including your new partial), so I think we're good to go there.

jayzes:
@marlabrizel do you agree?

PR #272, opened 2018-04-17
Users should be able to upload a profile photo for their accounts, issue #226

### LINKED ISSUE #226: Users should be able to upload a profile photo for their accounts
We can use the account edit page for this.

Might be good to add some variety of client-side validation too: https://stackoverflow.com/questions/3717793/javascript-file-upload-size-validation.

IlinDmitry:
Hello.
Some practice with carrierwave.

Fixes #226.


app/views/layouts/application.html.slim
```diff
@@ -27,6 +27,7 @@ html
       == yield
     = render partial: 'layouts/shared/site_footer'
     = javascript_pack_tag 'application'
+    = javascript_include_tag 'application'
```
jayzes: Rather than using the old asset pipeline, let's use the newer Webpack/Webpacker flow - we moved to that about a year ago.  You can see an example for the setup we have to bind components to a form [here](https://github.com/denverstartupweek/dsw-site/blob/master/app/javascript/components/autocompleter.js), [here](https://github.com/denverstartupweek/dsw-site/blob/d60aa6f7bd08d3e302c5b41013988fa3d0fc2cb7/app/javascript/packs/application.js#L24) and [here](https://github.com/denverstartupweek/dsw-site/blob/c3e77158732ab2a28e3e0b92f7f19f3c020fb798/app/views/registrations/new.html.slim#L38).

app/views/users/registrations/edit.html.slim
```diff
@@ -4,6 +4,11 @@ section.common
   .section-content.sm-centered
     = form_for(resource, as: resource_name, url: user_registration_path, html: { method: :put }) do |f|
 
+      = image_tag(resource.avatar.url(:thumb)) if resource.avatar?
+
+      = f.label :avatar, 'Avatar'
+      = f.file_field :avatar, onchange: 'showFileSize();'
```
jayzes: This can be the integration point for the `Bindable` component

app/views/users/registrations/new.html.slim
```diff
@@ -4,6 +4,11 @@ section.common
   .section-content.sm-centered
     = form_for(resource, as: resource_name, url: user_registration_path, honeypot: true) do |f|
 
+      = image_tag(resource.avatar.url(:thumb)) if resource.avatar?
+
+      = f.label :avatar, 'Avatar'
+      = f.file_field :avatar, onchange: 'showFileSize();'
```
jayzes: And here too.

config/initializers/carrierwave.rb
```diff
@@ -31,3 +31,27 @@
     config.storage = :file
   end
 end
+
+if Rails.env.test?
```
jayzes: I think this section may be unnecessary - it's the default pathing that Carrierwave sets up, and we default back to file storage when AWS is not set up (https://github.com/denverstartupweek/dsw-site/blob/master/config/initializers/carrierwave.rb#L28-L33)

spec/spec_helper.rb
```diff
@@ -31,4 +31,10 @@
   # the seed, which is printed after each run.
   #     --seed 1234
   config.order = "random"
+
+  config.after(:each) do
+    if Rails.env.test? || Rails.env.cucumber?
+      FileUtils.rm_rf(Dir["#{Rails.root}/spec/support/uploads"])
```
jayzes: Can we do this in a subdirectory of `tmp` for easy cleanup?

IlinDmitry:
Hello @jayzes ! Thank you for review. I learned something new about webpack and "bindable".

IlinDmitry:
Hi all.
Is something wrong with this pull request? Should i fix something?

PR #279, opened 2018-05-03
Fetch and group Helpscout articles by category

jayzes:
This lets us create categories for different FAQ pages and parcel them out accordingly.


app/helpers/application_helper.rb
```diff
@@ -2,8 +2,42 @@ module ApplicationHelper
 
   def process_with_liquid(content)
     context = {
-      'submission_close_date' => AnnualSchedule.current.cfp_close_at.in_time_zone(ActiveSupport::TimeZone['America/Denver']).at_end_of_day,
-      'voting_close_date' => AnnualSchedule.current.voting_close_at.in_time_zone(ActiveSupport::TimeZone['America/Denver']).at_end_of_day,
+      'submission_open_date' =>
+        AnnualSchedule.
```
marlabrizel: nitpicky perhaps, but can we put periods at the beginning of subsequent method calls? I think it's clearer that way. e.g.
```
AnnualSchedule
  .current
  .cfp_open_at
  ...
```

app/helpers/application_helper.rb
```diff
@@ -2,8 +2,42 @@ module ApplicationHelper
...
+          current.
+          cfp_open_at.
+          in_time_zone(ActiveSupport::TimeZone['America/Denver']).
+          at_beginning_of_day,
+      'submission_close_date' =>
+        AnnualSchedule.
```
marlabrizel: same here

app/helpers/application_helper.rb
```diff
@@ -2,8 +2,42 @@ module ApplicationHelper
...
+          current.
+          cfp_close_at.
+          in_time_zone(ActiveSupport::TimeZone['America/Denver']).
+          at_end_of_day,
+      'voting_open_date' =>
+        AnnualSchedule.
```
marlabrizel: and here

PR #280, opened 2018-05-04
Add coc_acknowledgement to admin permitted params

jayzes:
This was preventing admins from adding new sessions.

/cc @marlabrizel


jayzes:
Going to self review and merge since this is pretty small.

Issue #244, opened 2018-03-16
Fix display bug with "+" in FAQ sections


jayzes:
They should be on the same line as the title.

It's related to a margin-bottom on h6 elements that I added sometime last year. Need to make the selector more specific.

mmdberg:
Hi there! I’m a first-time contributor and was hoping to help out with this issue. I noticed nobody was assigned to it, but if there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

PR #282, opened 2018-05-08
Addressing issue #244

CharlesY712:
In faq.sass removed align items center
Changed padding to address all four sides
Added margin-bottom  in .panel and children

Before:
before

After:
after


jayzes:
Looks great - thanks!

PR #283, opened 2018-05-15
Create mechanism for session feedback

jayzes:
This implements #276.

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -1,0 +1,25 @@
+class FeedbackController < ApplicationController
+
+  before_action :authenticate_user!, only: %i(create)
+
+  def create
+    @feedback = current_user.feedback.new(feedback_params)
+
+    if @feedback.save
+      flash[:notice] = 'Thank you for submitting feedback!'
+      redirect_to schedule_path(@feedback.submission)
+    else
+      flash[:error] = 'there was a problem submitting feedback'
+      redirect_to schedule_path(@feedback.submission)
+    end
+  end
+
+  private
+
+  def feedback_params
+    params.require(:feedback).permit(:rating,
+                                     :comments,
+                                     :user_id,
+                                     :submission_id)
+  end
+end
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -1,17 +1,16 @@
 class FeedbackController < ApplicationController
 
-  before_action :authenticate_user!, only: %i(create)
+  before_action :authenticate_user!, only: :create
 
   def create
     @feedback = current_user.feedback.new(feedback_params)
 
     if @feedback.save
       flash[:notice] = 'Thank you for submitting feedback!'
-      redirect_to schedule_path(@feedback.submission)
     else
       flash[:error] = 'there was a problem submitting feedback'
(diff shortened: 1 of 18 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  before_action :authenticate_user!, only: :create
+
+  def create
+    @feedback = current_user.feedback.new(feedback_params)
+
+    if @feedback.save
+      flash[:notice] = 'Thank you for submitting feedback!'
+    else
+      flash[:error] = 'there was a problem submitting feedback'
```
marlabrizel: We should clean this message up to not be my hacky placeholder :) Maybe `There was a problem submitting feedback. Please try again later or email info@denverstartupweek.org with the session name and your comments` - something like that?

app/controllers/schedules_controller.rb
```diff
@@ -109,7 +109,7 @@ def destroy
   def current_day_or_default
     if AnnualSchedule.in_week?
       zone = ActiveSupport::TimeZone['America/Denver']
-      day_index = ((zone.now.at_beginning_of_day - AnnualSchedule.week_start_at.at_beginning_of_day.in_time_zone(zone)) / 1.day).ceil + 2
+      day_index = ((zone.now.at_beginning_of_day - AnnualSchedule.current.week_start_at.at_beginning_of_day.in_time_zone(zone)) / 1.day).ceil + 2
```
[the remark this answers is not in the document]
marlabrizel: if we care about hound, we could maybe split up each operation to its own line - this is kind of hard to read but up to you

app/models/submission.rb
```diff
@@ -139,6 +140,10 @@ def editable?
     created? || open_for_voting? || accepted? || confirmed? || venue_confirmed?
   end
 
+  def open_for_feedback?
```
marlabrizel: can we use the `in_or_post_week` helper in `schedule_helper.rb` to replace some of this logic? we might need to edit that to scope to current year but I think that's what that helper was created for

jayzes: The tough thing is that `in_or_post_week` is in a view helper context, so it's not callable from here. I did think about where this ought to live - I could have made it another view helper function, but since it always relates to an individual submission (with the past years thing) I thought it felt slightly awkward to have as a view helper. `in_or_post_week` is also designed for just the current cycle, whereas this lets you provide feedback anytime after the week has started for that submission's year and on into the future.

app/controllers/schedules_controller.rb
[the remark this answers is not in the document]
jayzes: I'm not too concerned about it for now - at some point I want to refactor `AnnualSchedule` to eliminate all of the date math gymnastics we have peppered throughout the codebase. Until then I'm not going to sweat a bit of localized hackiness. At least it's consistent, haha.

app/models/submission.rb
marlabrizel: That makes sense. 

Do we want to cut off the feedback window? Seems like 3 years out, for example, feedback is going to be distorted by time. But I suppose we can also filter that out by timestamp. 

Anyway I'm fine to leave this as is- after all, perfect is the enemy of good

PR #298, opened 2018-07-05
Remove day_of_week from schedule day chiclet on mobile

SageVanGogt:
Added a new media query that sets the day_of_week element display to none when screen size is less than 600px.


jayzes:
Looks good to me - thanks!

PR #299, opened 2018-07-18
First attempt at attendee goals tracking at registration

jayzes:


jayzes:
Still could use:

- Eyes on the list of goals to start 2018 with
- An admin

PR #356, opened 2019-04-03
Venue Selector

dummey:
This MR adds in the ability for a venue host to create a venue via their dashboard. And for admins to assign submissions to venue time slots. 

The goal is to enable the venue host and track leads to create, manage, and assign everything needed to deal with venues and scheduling through the DSW site.


jayzes:
Thanks @dummey ! I'll land this into an integration branch and take a look at the hound warnings and tests.

PR #368, opened 2019-05-17
Example of how to structure code and example session cards

scpaye506:
@andylushman @chrisman 

Thoughts on this approach?

Mostly a component based approach to building on the templates/partials and then using a component based css style guide called SUIT

https://suitcss.github.io/

can see the components on the home page if you scroll down


scpaye506:
@andylushman SUIT not BEM

https://suitcss.github.io/

They are similar though.

PR #370, opened 2019-05-20
Secondary nav ctas

scpaye506:


app/assets/stylesheets/redesign/_config.sass
```diff
@@ -1,3 +1,10 @@
+// FONT COLORS
+$primary-font-light: rgb(95, 95, 95)
+$primary-font-dark: rgb(3, 92, 107)
+
+// BORDER COLORS
+$primary-border-light: rgb(151, 151, 151)
+
```
andylushman: Glad to see we have somewhere to add global vars.

app/assets/stylesheets/redesign/components/nav.sass
```diff
@@ -1,30 +1,30 @@
...
+//     border-bottom: 5px solid #fff
+//     color: $teal-light
+//     cursor: pointer
+//     &:last-child
+//       margin-right: 0
+//     &:hover, &:active, &.active
+//       border-bottom: 5px solid $teal-light
+//       color: $teal-dark
+//   &.section-nav
+//     margin-bottom: 30px
```
andylushman: In general, are we keeping most things and commenting them out instead of deleting them?

app/assets/stylesheets/redesign/components/nav.sass
```diff
@@ -77,7 +77,7 @@ nav
...
 
 
-@media(min-width: 375px)
-  nav
-    a
-      margin-right: 30px
+// @media(min-width: 375px)
+//   nav
+//     a
+//       margin-right: 30px
```
andylushman: I added a question for Geoff about what the breakpoints are for the new screens.

app/assets/stylesheets/redesign/components/nav.sass
scpaye506: i haven't thought about it too much yet. My initial thoughts on this were I didn't want to delete something we could possibly re-use later, but it was currently impacting my styles bc of how broad the specificity is

PR #371, opened 2019-05-20
Footer Refactor

andylushman:
- Renamed file `_site_footer.html.slim` to `_footer.html.slim`
- Refactored `_footer.html.slim` and `footer.sass`
     - Added SUIT naming conventions
     - Added CSS to make font-awesome social icons round

Footer is flush with the edge of the cards located above it when at full screen and aligned on the left when in mobile. It is `justify-content: space-between` and adjusts accordingly in between the screens.

I tried removing `page_footer.html.slim` from the `index.html.slim` but it causes some of the test to fail. The issue is that they are looking for the email address that is included in the `page_footer.html.slim`. Instead of modifying the test I added the `page_footer.html.slim` back in. Happy to remove it and the test if we feel like that's a better solution.


app/views/layouts/shared/_footer.html.slim
```diff
@@ -0,0 +1,35 @@
+footer.Footer
+  .Footer-blocks
+    .Footer-block
+      .Footer-company-container
```
scpaye506: It doesn't look like this `.Footer-company-container` div is containing anything. Maybe can remove it and the styles for it?

app/views/layouts/shared/_footer.html.slim
```diff
@@ -0,0 +1,35 @@
...
+      div.Footer-link= link_to 'Attendees', page_path(page: 'attendees')
+      div.Footer-link= link_to 'Podcast', page_path(page: 'podcast')
+      div.Footer-link= link_to 'Press', page_path(page: 'press')
+      div.Footer-link= link_to 'Code of Conduct', page_path(page: 'code-of-conduct')
+
+    .Footer-block
+      h6 Denver Startup Week #{Date.today.year}
+      h5 &copy;Copyright #{Date.today.year}. All rights reserved.
+      .Footer-social
+        a href='https://twitter.com/denstartupweek' target='_blank'
```
scpaye506: thoughts on making a Font Awesome component?

app/assets/stylesheets/redesign/views/footers.sass
```diff
@@ -1,74 +1,66 @@
...
+      height: 15px
+      color: #fff
 
-.footer-blocks
+.Footer-blocks
   display: flex
   flex-direction: column
   margin: 0 auto
   padding: 40px 20px
   max-width: 1400px
```
scpaye506: i noticed in the sass config there is a var for `max-width` we probably want to look into using.

app/assets/stylesheets/redesign/views/footers.sass
```diff
@@ -1,74 +1,66 @@
...
     text-transform: uppercase
     white-space: nowrap
-  .social
+  h5
+    font-size: 12px;
+  .Footer-social
     display: inline-block
     vertical-align: middle
     margin-top: 20px
+    .fa-circle-thin
```
scpaye506: if you pull the font awesome stuff into it's own component then we could more easily follow the SUIT naming convention for this class name (although even without pulling it into it's own component you could probably update the naming convention here)

app/assets/stylesheets/redesign/views/footers.sass
```diff
@@ -1,74 +1,66 @@
...
-    max-width: 600px
-    padding: 0
-    margin: 0 auto
-  h1
-    color: $copy-color-bold
-
-footer.site-footer
-  background: $backfill-grey
+.Footer
   h6
```
scpaye506: ideally if we can, let's avoid using tag selectors (or refactor them out to classes if we can)

app/views/layouts/shared/_footer.html.slim
andylushman: Good catch! Just removed it.

app/assets/stylesheets/redesign/views/footers.sass
andylushman: Great! Just added `$max-breakpoint`

app/assets/stylesheets/redesign/views/footers.sass
andylushman: Got it. Added a class for `title` and `subtitle`

app/assets/stylesheets/redesign/views/footers.sass
andylushman: Took a quick look through the site and didn't see any other places where font awesome was being used. I did change this name to `.Footer-fa-circle`. If you think it's still valuable to create a component for font awesome, let me know and I will take care of it.

app/views/layouts/shared/_footer.html.slim
andylushman: If you this its valuable I can create a component for it. I took a quick look through the site and It looks like its only being used for the social icons on this page.

app/views/layouts/shared/_footer.html.slim
```diff
@@ -0,0 +1,35 @@
...
+      - if signed_in? && current_user.is_admin?
+        div.Footer-link= link_to 'Admin', admin_root_path
+      div.Footer-link= link_to 'Attendees', page_path(page: 'attendees')
+      div.Footer-link= link_to 'Podcast', page_path(page: 'podcast')
+      div.Footer-link= link_to 'Press', page_path(page: 'press')
+      div.Footer-link= link_to 'Code of Conduct', page_path(page: 'code-of-conduct')
+
+    .Footer-block
+      h6 Denver Startup Week #{Date.today.year}
+      h5 &copy;Copyright #{Date.today.year}. All rights reserved.
```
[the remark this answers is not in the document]
andylushman: Yeah, I removed the `h6` and `h4` tags and replaced them with `div` when I added a class.

app/views/layouts/shared/_footer.html.slim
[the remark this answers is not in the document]
andylushman: Nice! I like it. Added a `nav` and `ul` tag and put in some `aria-labels`

app/views/layouts/shared/_footer.html.slim
```diff
@@ -0,0 +1,35 @@
...
+      .Footer-social
```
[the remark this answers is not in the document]
andylushman: Nice! I like it! Added `nav` and `ul` tags.

app/views/layouts/shared/_footer.html.slim
```diff
@@ -1,35 +1,41 @@
...
-  .Footer-blocks
-    .Footer-block
-      .Footer-company-container
-      - if signed_in? && current_user.is_admin?
-        div.Footer-link= link_to 'Admin', admin_root_path
-      div.Footer-link= link_to 'Attendees', page_path(page: 'attendees')
-      div.Footer-link= link_to 'Podcast', page_path(page: 'podcast')
-      div.Footer-link= link_to 'Press', page_path(page: 'press')
-      div.Footer-link= link_to 'Code of Conduct', page_path(page: 'code-of-conduct')
+  nav aria-label="Footer items"
```
scpaye506: i think this nav is a bit too high level. I don't think it needs to wrap the footer blocks. just the nav lists in each block

app/views/layouts/shared/_footer.html.slim
```diff
@@ -1,35 +1,41 @@
...
-        a href='https://www.youtube.com/c/denverstartupweek' target='_blank'
-          span class="fa-stack"
-            i class="fa fa-circle-thin fa-stack-2x"
-            i class="fab fa-2x fa-youtube fa-stack-1x"
-        a href='https://medium.com/denver-startup-week' target='_blank'
-          span class="fa-stack"
-           i class="fa fa-circle-thin fa-stack-2x"
-           i class="fab fa-2x fa-medium-m fa-stack-1x"
\ No newline at end of file
+      li.Footer-block
```
scpaye506: so for example, this doesn't need to be an `li`

app/views/layouts/shared/_footer.html.slim
```diff
@@ -1,35 +1,41 @@
...
-      .Footer-company-container
-      - if signed_in? && current_user.is_admin?
-        div.Footer-link= link_to 'Admin', admin_root_path
-      div.Footer-link= link_to 'Attendees', page_path(page: 'attendees')
-      div.Footer-link= link_to 'Podcast', page_path(page: 'podcast')
-      div.Footer-link= link_to 'Press', page_path(page: 'press')
-      div.Footer-link= link_to 'Code of Conduct', page_path(page: 'code-of-conduct')
+  nav aria-label="Footer items"
+    ul.Footer-blocks
+      li.Footer-block aria-label="Page links"
```
scpaye506: Sorry, I think maybe I put my comment on the wrong line, but this `.Footer-block` should probably be the `ul` and all of the `div.Footer-link`'s should be the `li`'s

app/views/layouts/shared/_footer.html.slim
andylushman: Makes sense.

app/views/layouts/shared/_footer.html.slim
andylushman: Perfect. That's what I was thinking. Making the changes now.

app/views/layouts/shared/_footer.html.slim
andylushman: Just pushed up the changes 👍

PR #380, opened 2019-05-23
Program layout

scpaye506:
needs https://github.com/denverstartupweek/dsw-site/pull/379 reviewed/merged first


app/data/programs.yml
```diff
@@ -1,21 +1,21 @@
...
-    img_url: "basecamp/slide10_large.jpg"
-    button_page_url: "tracks"
+    img_url: 'basecamp/slide10_large.jpg'
+    button_page_url: 'tracks'
   - title: 'key events for 2019'
     description: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque dictum massa non magna ornare, et lacinia nibh semper. Curabitur eu varius justo. In cursus sodales arcu et mattis.'
-    img_url: "basecamp/slide10_large.jpg"
-    button_page_url: "tracks"
+    img_url: 'basecamp/slide10_large.jpg'
+    button_page_url: 'tracks'
```
andylushman: Was this a `prettier` thing or are we using single quotes for these types of files?

scpaye506: ah, i think it was an auto format thing

scpaye506:
> What's your thoughts on putting `padding` on the `main` tag so that we don't have to do it for each page/component?

@andylushman It would be nice but adding `padding` to the `main` tag breaks the way i've done those color triangles behind the featured sessions and articles on the home page

PR #382, opened 2019-05-28
Tracks layout

scpaye506:


app/assets/stylesheets/redesign/core/track-colors.sass
```diff
@@ -22,4 +22,78 @@
...
+  border-color: $people-track
+
+.founder-border-color
+  border-color: $founder-track
+
+.maker-border-color
+  border-color: $maker-track
+
+.growth-border-color
+  border-color: $growth-track
```
andylushman: I like how these colors are all now stored as vars 👍

PR #383, opened 2019-05-29
Desktop and Mobile menu

scpaye506:
created a separate social media nav component

the menu is mostly complete for both desktop and mobile, it's just missing the header stuff right now.

This includes the logic for nested sub-menus and for knowing when to open a nested sub-menu based on the current route

It does not include every route that will be part of the menu yet (or sub menus), and it has the start to including some of the logic for what menu items to show based on what timeframe we are in (copied from the current nav menu), but that will need some more refinement


app/views/components/_menu.html.slim
```diff
@@ -0,0 +1,24 @@
+.Menu#menu-js(data-bindable='menu' data-starting_nav="#{menu_starting_nav}")
```
scpaye506: @chrisman @andylushman if you want to see the menu (since there is no button to open it right now) add the class `.is-open` like this `.Menu.is-open`

app/assets/stylesheets/redesign/components/menu.sass
```diff
@@ -0,0 +1,141 @@
...
+
+  .Menu-item-link.has-nested
+    display: none
+
+  .Menu-nav
+    position: absolute
+    top: 0
+    bottom: 0
+    width: 100%
+    transition: all $transition-time linear
```
andylushman: Don't think I've seen this before.

app/javascript/components/menu.js
```diff
@@ -0,0 +1,98 @@
...
+    $(document)
+      .find(MENU_PRIMARY_NAV_ID)
+      .addClass('is-closed');
+
+    $(document)
+      .find(`#${nav}-${MENU_SECONDAY_NAV_SUFFIX_ID}`)
+      .addClass('is-open');
+  }
+
+  closeSecondaryNavClick(event) {
```
andylushman: Nice to read JS again. Except maybe for the jQuery.

app/views/components/_menu.html.slim
andylushman: 👍

app/assets/stylesheets/redesign/components/menu.sass
```diff
@@ -0,0 +1,141 @@
...
+
+.Menu
+  position: absolute
+  top: 0
+  left: 0
+  right: 0
+  bottom: 0
+  background: $teal-dark
+  color: white
+  z-index: 11
```
chrisman: z-index inflation always feels weird to me, but I don't know what to do about it. I assume this 11 is to stack over the 10 on view/header?

app/helpers/route_helper.rb
```diff
@@ -1,9 +1,91 @@
 module RouteHelper
 
+  def main_menu
+    menu = [home_route, main_program_route, sponsers_route, get_involved_route]
+    if AnnualSchedule.cfp_open?
```
chrisman: 1. can this block be changed so that the routes and the schedules are more consistent? e.g. `if AnnualSchedule.submissions_open? menu.push submissions_route`

2. I don't know what `post_week` is. Based on the surrounding lines, should it be something like `schedule_open` instead?

app/javascript/components/menu.js
```diff
@@ -0,0 +1,98 @@
...
+
+const BODY_MENU_CLASS = 'Body-menu-open';
+const MENU_OPEN_CLASS = 'is-open';
+const MENU_ID = '#menu-js';
+const MENU_CLOSE_ID = '#menu-close-button-js';
+const MENU_OPEN_ID = '#menu-open-button-js';
+const MENU_OPEN_SECONDARY_NAV_ID = '#open-secondary-nav-js';
+const MENU_CLOSE_SECONDARY_NAV_ID = '#close-secondary-nav-js';
+const MENU_PRIMARY_NAV_ID = '#primary-nav-js';
+const MENU_SECONDAY_NAV_SUFFIX_ID = 'secondary-nav-js';
```
chrisman: typo: "secondary"

autocomplete was not your friend here, but I think it only shows up 3 total times.

app/views/layouts/shared/_footer.html.slim
```diff
@@ -9,33 +9,7 @@ footer.Footer
...
-              span class="fa-stack"
-                i class="fa Footer-fa-circle fa-stack-2x"
-                i class="fab Footer-social-icon fa-2x fa-youtube fa-stack-1x"
-          li
-            a.Footer-social-link aria-label="Medium" href='https://medium.com/denver-startup-week' target='_blank'
-              span class="fa-stack"
-                i class="fa Footer-fa-circle fa-stack-2x"
-                i class="fab Footer-social-icon fa-2x fa-medium-m fa-stack-1x"
\ No newline at end of file
+      = render 'components/social_media_nav'
```
chrisman: much cleaner. good job refactoring this out.

app/assets/stylesheets/redesign/components/menu.sass
scpaye506: yeah exactly. We can probably pull these numbers back down into the single digits if we want, didn't want to mess with the header one though while you were re-building it.

app/javascript/components/menu.js
scpaye506: good catch, thanks!

app/helpers/route_helper.rb
scpaye506: I'm not sure i totally follow your first point, can you elaborate a bit?

I'm not really sure what `post_week` is either, I took all of this logic directly from their current menu. I wasn't planning on really modifying it yet, just moving it out of the view into a helper method.

PR #384, opened 2019-05-29
Fix margins and fonts between home hero, secondary callouts, and feat…

scpaye506:
…ured


scpaye506:
> Noticed that we've been using `max-width` to change mobile designs. Are we not doing mobile first?

Their current css is a mix of both, I don't really have a strong preference one way or another with it.

PR #385, opened 2019-05-29
Track detail template

andylushman:
- Created `track_details.html.slim` which
     - renders the `track_details_hero` partial
     -  renders the `individual_track` partial
     - renders the `primary_callouts_nav`

Once we decide on routing for tracks and clusters, these files will need to be changed accordingly. For example, the `individual_track` could take params from the route, or we may end up building files for each individual track .Currently, if you go to `program/track_details` you can view the styling and layout. It is default to the `designer` track.


app/views/layouts/shared/_track_details_hero.html.slim
```diff
@@ -0,0 +1,10 @@
+.TrackDetailsHero
```
scpaye506: this type of hero will be used with the cluster page as well, can we make it a bit more generic so it can easily be re-used?

app/helpers/route_helper.rb
```diff
@@ -18,6 +18,13 @@ def tracks_route
     }
   end
 
+  def tracks_route_track_details
```
scpaye506: there won't be a route with the path `program/track_details`, it'll probably look something more like
```
def track_detail_route(name)
  {
     path: "program/tracks/#{name}",
     label: name
  }
end
```

scpaye506: it doesn't actually look like you're using this anywhere though

app/views/components/_track_details_button.html.slim
```diff
@@ -0,0 +1,3 @@
+button.TrackDetailsButton
```
scpaye506: same with this being used with other pages like the cluster detail pages. I'm not sure this needs to be it's own component, maybe just part of the hero component, but if you keep it as it's own let's make it more generic

app/views/layouts/shared/_individual_track.html.slim
```diff
@@ -0,0 +1,11 @@
+section.Tracks
```
scpaye506: this feels like it could probably be named something like `_track_list.html.slim` to give it a name with a little more context (all of your classes almost follow that pattern in this file anyways)

scpaye506:
looks like there is a layout issue with the first card once they start stacking, like some padding or margin left

app/views/layouts/shared/_track_details_hero.html.slim
andylushman: good idea. I changed it to `SubProgramDetailsHero` I didn't love `SubProgram` but I was trying to think of a name that could work for both `tracks` and `clusters`.

app/views/components/_track_details_button.html.slim
andylushman: good call. Made it a part of the hero component

app/views/layouts/shared/_individual_track.html.slim
andylushman: good call. Just renamed it and the corresponding `sass` file.

andylushman:
> looks like there is a layout issue with the first card once they start stacking, like some padding or margin left

This was being caused but adding unnecessary `li` to the `ul` in `three_card_list` I fixed it with this...
```
  = render 'components/three_card_list', attributes: { 'aria-label': 'Designer sessions and atricles' } do
      - tracks[:features].each do |track|
        - if track[:type] == 'session' && track[:track] == 'designer'
          = render 'components/three_card_list_item' do
            = render 'components/session_card', session: track
        - if track[:type] == 'article' && track[:tracks].include?('designer')
          = render 'components/three_card_list_item' do
            = render 'components/article_card', article: track
```

app/views/site/program/track_details.html.slim
```diff
@@ -0,0 +1,5 @@
+= render partial: 'layouts/shared/sub_program_details_hero', locals: { title: "Designer", subProgram: "Tracks", backUrl: "tracks", btnSessions: "Track", backgroundColor: "rgb(187, 221, 122)", textColor: "rgb(97, 107, 34)" }
+/ div TRACK DETAIL PAGE FOR #{@track[:name]}
```
andylushman: @scpaye506, still not totally sure about how to use this here.

andylushman:
@scpaye506, made some updates to make the code more generic and the hero should be reusable for the clusters page. Also, added some routing code that I noticed in your `clusters` PR. Still not totally sure how to get it wired up. Everything else for the story should be completed except for `A user can click on a track event to view the session’s details.` Which sounds like from our conservation this morning, will have to wait. Once I figure out how to do the routing, I will have to make some changes to `_track_list.html.slim`

app/assets/stylesheets/redesign/layouts/sub_program_details_hero.sass
```diff
@@ -0,0 +1,71 @@
+@import ../config
+
+.SubProgramDetailsHero
+  // background: rgb(187, 221, 122)
```
andylushman: Remove colors from this file, if they are staying in `_sub_program_details_hero.html.slim `

app/views/site/program/track_details.html.slim
andylushman: Thanks for your help this morning. Understanding it now.

app/assets/stylesheets/redesign/layouts/sub_program_details_hero.sass
andylushman: Removed the colors from sass file.

app/views/site/program/track_details.html.slim
```diff
@@ -0,0 +1,8 @@
+- color = track_color("#{@track[:name]}")
+- backgroundColor = color[:backgroundColor]
+- textColor = color[:textColor]
+
+= render partial: 'layouts/shared/sub_program_details_hero', locals: { title: "#{@track[:name]}", subProgram: "Tracks", backUrl: "/program/tracks", btnSessions: "Track", backgroundColor: backgroundColor, textColor: textColor }
+= render partial: 'layouts/shared/track_list', locals: { tracks: @featured, trackName: "#{@track[:name].downcase}" }
```
scpaye506: i feel like the naming of things here might be a bit off, and something we could make more intuitive. For example the track_list is a list of sessions and articles, not a list of tracks. so maybe the local var that is passed to it should be something like `sessions_and_articles` instead of `tracks`

also rails convention for variable names is snake_case not camelCase, so `trackName` should be `track_name` and any of the other variables that you have camelCased

app/helpers/sub_program_helper.rb
```diff
@@ -0,0 +1,26 @@
+module SubProgramHelper
```
scpaye506: instead of having this helper, is there anyway we can use the track classes i created in `assets/stylesheets/redesign/core/track-colors.sass`?

app/controllers/tracks_controller.rb
```diff
@@ -0,0 +1,13 @@
+class TracksController < ApplicationController
+  def show
+    @secondary_nav = HashWithIndifferentAccess.new(YAML.safe_load(
+      File.read(File.expand_path("../data/secondary_nav.yml",  __dir__))))
+
+    @featured = HashWithIndifferentAccess.new(YAML.safe_load(
+      File.read(File.expand_path("../data/featured.yml",  __dir__))))
+    
+    @track = Track.find_by(name: params[:name])
+      render template: "site/program/track_details"
```
scpaye506: can we make it so it pulls a template from this path `site/program/tracks/show`?

andylushman: For sure. Done!

app/views/site/program/track_details.html.slim
andylushman: Made the switch to `sessions_and_articles` and changed the rest of the vars to snake_case. Will use snake_case in the future when writing in rails.

app/helpers/sub_program_helper.rb
andylushman: as per our conservation, I added the following classes for each track/color.
file: `track_colors.sass`
```
.#{color}-font-on-#{color}
  color: ${color}-dark

.#{color}-background
  background-color: $color-light

.#{color}-border-on-#{color}-background
  border-color: #{color}-light
```

file: `_config.sass`
for the most part, I was able to use the colors that where already there. I did however, have to add a couple of colors for orange and gold.

app/assets/stylesheets/redesign/_config.sass
```diff
@@ -17,9 +17,12 @@ $teal-very-light: #91ECEF
 
 $orange-dark: #844212
 $orange-light: #ED853E
+$orange-background: #EC833E
```
scpaye506: just one small change - i think we can probably just replace make `$orange-light` this new `EC833E` hex instead of creating a sole color var with a `-background` suffix.

Then good to merge!

andylushman: Sounds good! Just made the change.

PR #386, opened 2019-05-29
Initial route, controller, view setup

scpaye506:
this is how we'll do the routing for pages like clusters and tracks.

So the route in this example would be something like `{root_url}/program/clusters/:name`

The show view will then load the track/cluster/whatever based on the passed name param


config/routes.rb
```diff
@@ -71,6 +71,10 @@
   resources :general_inquiries, only: :create
   resources :sponsor_signups, only: :create
 
+  scope :program do
```
chrisman: nice. is there a reason you used `scope` here instead of `namespace`?

scpaye506: `namespace` seemed to want the controller to be nested under a `program` module and didn't think it was worth re-structuring the controller directory for that.

scpaye506:
@andylushman @chrisman i've updated this pr a bit since your last review to include the actual cluster detail page layout based on the work Andy did on the track detail page. If you want to re-review it

PR #387, opened 2019-05-29
Set up article data model & admin

jayzes:


scpaye506:
@jayzes I realize this is still WIP, but figure I'll ask/comment now - looking at the designs, it looks like there is a need for the model to support hero and thumbnail images (on the article detail page and then on the thumbnail on the article card). It also looks like articles can/should be associated to a track

PR #390, opened 2019-05-31
Link content cards to corresponding page

andylushman:
Cards on the `tracks` page now link to each individual track. Used the `capitalized` method  so that the url is correct. Example: `/developer` becomes `/Developer`.


app/views/components/_content_card.html.slim
```diff
@@ -11,4 +11,4 @@ div.ContentCard
       h3.ContentCard-title(class="#{card[:track]}-font") #{card[:title]}
     p.ContentCard-description #{card[:description]}
     .ContentCard-button
-      = render 'components/content_button', title: "Explore the #{title}", button_page_url: "#{card[:button_page_url]}"
\ No newline at end of file
+      = render 'components/content_button', title: "Explore the #{title}", track_title: "#{card[:track].capitalize}"
```
scpaye506: this needs to be more dynamic, this content card is used by more than just tracks

andylushman: I was thinking of two ways of doing this.
1. (current in this PR): when calling `card_content` we pass down the beginning of the url. For example in this case we can use `program/tracks`. Then in `content_card` we add `#{card[:page_url_title]}` to it. This would have to be added to the card object.

2. We just use #{card[:page_url]} and add the page url to the card. Although, this would probably not be as flexible.

scpaye506: I'm sort of thinking we get rid of the `_card_content.html.slim` file so that we can pass the url easier

andylushman: Removed `_card_content.html.slim` and everything should be working. However, I left the final `page_url` incomplete on both the `program` and `clusters` pages, since I wasn't sure what we are using. We had talked about title but that currently wouldn't work for the `tracks` page.

PR #402, opened 2019-06-05
Session detail page

rflorence:
This does not include the list of attendees or a working Directions button, but it does include a sass mixin to create those white buttons with rounded teal borders in dynamic sizes.


app/assets/stylesheets/redesign/components/buttons.sass
```diff
@@ -79,3 +79,18 @@ a.btn.gray
     background-position: 0 0
     background-repeat: no-repeat
     background-size: cover
+
+=white-button($width, $height)
```
scpaye506: oh cool, i didn't know this `=` and `+` syntax

scpaye506: what do you think about adding a mixins file and importing that into config?

scpaye506:
one thing we've been doing is where we are refactoring the html, we are also refactoring the css to follow SUIT conventions

For the layout it looks like everything is center aligned, but in the designs it looks like everything is left aligned with the primary callouts

app/assets/stylesheets/redesign/components/buttons.sass
rflorence: I didn't know until I started this project that .scss and .sass are different formats. The [guide](https://sass-lang.com/guide) has tabs for the different syntaxes.

rflorence: Fine with me! There are too many different stylesheets for me to understand and form an opinion on them yet, but having them all already included in all of the files would be nice.

rflorence:
I've refactored css to SUIT now and moved the content to a 5% padding rather than centered max-width. The content won't be full width once I add in the attendees list, but that's in a separate story.

app/assets/stylesheets/redesign/views/session_detail.sass
```diff
@@ -1,200 +1,126 @@
...
+  margin-bottom: 20px
+  font-size: 28px
+  color: $primary-font-dark
+  text-transform: uppercase;
+.SessionDetail-stats-container
+  display: flex
+  flex-direction: row
+.SessionDetail-date, .SessionDetail-time, .SessionDetail-attendee-count
+  text-transform: uppercase
+  font-family: "Montserrat-Regular"
```
scpaye506: looks like this should be `Montserrat-Medium`

also, best if we use the `font-family-mixin(Montserrat-Medium)` so that it uses the fallback font-family

PR #406, opened 2019-06-05
Logged in header

scpaye506:


app/views/layouts/shared/_header.html.slim
```diff
@@ -4,7 +4,11 @@ header#site-header-js.Header
       img.Header-image(src="#{image_url('redesign/logos/dsw-grey.svg')}")
       img.Header-image-menu(src="#{image_url('redesign/logos/dsw-white.svg')}")
     .Header-actions
-      a.Header-signup(href="/") Sign Up / Sign In
+      - if signed_in?
+        a href="#{dashboard_path}"
+          = render partial: 'components/name_with_avatar', locals: { name: current_user.name, initials: 'SP', align: 'left' }
```
scpaye506: @rflorence you can probably use this same partial for the session users list thing

i need to add a method to get a users initials

app/models/user.rb
```diff
@@ -48,4 +48,8 @@ def registered?
   def flipper_id
     to_global_id.to_s
   end
+
+  def initials
+    name.split(" ").map { |n| n[0,1] }.join
```
rflorence: This is why you should always use separate `first_name` and `last_name` fields!

scpaye506: for sure!

PR #411, opened 2019-06-07
"added to schedule" text styling

rflorence:
Show "added to schedule" as static text instead of a button after you've added a session to your schedule.

It still needs an asset for a checkmark in order to match the design, but this is better than keeping the button!


scpaye506:
@jayzes so should we just disregard this design https://projects.invisionapp.com/d/?origin=v7#/console/17579255/366084166/preview

and revert to having the `Remove from Schedule` button?

PR #428, opened 2019-06-17
About page

andylushman:
* Added about Page
* Added `about.yml` file
* Added About to nav menu
* Created routes for sub menu
* Added placeholders for 
     * Assets page
     * Faq page
     * Team page
* Modified `content_card` so that it now also takes `arrays` in addition to `strings` for the description.

* To get the project up and running I had to migrate the db which caused a changed to the `structure.sql`


app/views/site/about.html.slim
```diff
@@ -0,0 +1,12 @@
+= render layout: 'layouts/shared/secondary_hero', locals: { title: 'About', sub_title: 'By the community, for the community', description: 'Bringing Denver Startup Week to life takes the support, skills, and drive of an entire community: entrepreneurs, investors, civic leaders, educators, business pioneers, and more.' } do
+  = render 'layouts/shared/about_secondary_hero_nav'
+
+div.Program-list-wrapper
```
scpaye506: I know this is the same layout, but better if we don't re-use class names here

app/views/site/about.html.slim
```diff
@@ -0,0 +1,12 @@
...
+  = render 'components/content_card_list', attributes: { 'aria-label': "About page" }  do
+    - about_static_data[:items].each do |card|
+      = render 'components/content_card_list_item' do
+        = render 'components/content_card', card: card, title: "#{card[:title]}", page_url: "about/#{card[:button_page_url]}"
+
+= render layout: 'layouts/shared/layout_wrapper' do
+  div.Program-primary-callouts
```
scpaye506: same here with the class name

db/structure.sql
```diff
@@ -9,20 +9,6 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
---
```
scpaye506: i know it did this automatically, but go ahead and revert the change in the file

andylushman: Removed that changes.

app/views/site/about.html.slim
andylushman: Added the classes to `about.sass`, which already existed"

app/views/site/about.html.slim
andylushman: Added to `about.sass`

db/structure.sql
```diff
@@ -9,34 +9,6 @@ SET xmloption = content;
...
```
scpaye506: there should be no diff for this file

andylushman: Just fixed it 👍

PR #429, opened 2019-06-17
Profile

scpaye506:
First part of the dashboard page

Still need to add the update profile page in a separate PR

This breaks all of the dummy article and session lists, so I've made it just return an empty array for now until we get real data


app/assets/stylesheets/redesign/components/toggle.sass
```diff
@@ -0,0 +1,43 @@
...
+  display: inline-block
+  width: 48px
+  height: 25px
+
+.Toggle input
+  opacity: 0
+  width: 0
+  height: 0
+
+.Toggle-slider
```
andylushman: Looks like the toggle slider is backwards from what is in the mock. For example, in the mock its teal when on the left vs on the right, as it is here.

app/views/dashboards/show.html.slim
```diff
@@ -1,40 +1,122 @@
...
+          li.Dashboard-card-list-item
+            = render partial: 'components/article_card', locals: { article: article }
+    - if @articles.blank?
+      .Dashboard-empty-message You have not submitted any content for this year.
+
+  section.Dashboard-section
+    .Dashboard-header-wrapper
+      h1.Dashboard-header My Sessions
+      - if AnnualSchedule.cfp_open?
+        a.Dashboard-header-action href="#{new_submission_path}"
```
andylushman: This currently throws an error but I'm guessing it something that needs to be taken care of by Jay?

app/assets/stylesheets/redesign/components/toggle.sass
scpaye506: yeah, good notice. It's pretty standard for the toggle to be off when it's left and on when it's right, which is opposite of the designs. Was going to ask Geoff about it at some point. Not too worried about it not matching the design atm though.

app/views/dashboards/show.html.slim
scpaye506: what error is it throwing?

app/assets/stylesheets/redesign/components/toggle.sass
andylushman: Agreed. My first thought was that the design was off.

app/views/dashboards/show.html.slim
andylushman: never-mind. It was because I didn't have redis running. Working now 👍

PR #431, opened 2019-06-18
Footer primary callouts

scpaye506:
moving the primary callouts into the footer instead of putting them at the bottom on every page.

Have a way to disable it on the home page


scpaye506:
@andylushman i'll remove it from the about page as part of this PR

PR #434, opened 2019-06-18
Team page

andylushman:
Added team page by creating:
* `team.html.slim`
* `team_member_card`
* `team_member_list`
* `static data with team_members.yml`


app/data/team_members.yml
```diff
@@ -0,0 +1,16 @@
+members:
```
scpaye506: we don't want to use static data for this. we want to get the team members the same way the old team page gets them.

scpaye506:
I think we need to add a redirect from the old team path to the new one

andylushman:
Just added a redirect: `get "/get-involved/team", to: redirect("/about/team")`

app/data/team_members.yml
andylushman: I think I have it done correctly now but it's taking me some time to figure out how to add the seed data.

andylushman: I think I have it done correctly now but it's taking me some time to figure out how to add the seed data. I added a `User` to `seeds.rb` and ran `rake db:seed` but the database doesn't get populated.

app/views/components/_team_member_card.html.slim
```diff
@@ -0,0 +1,5 @@
+= render 'components/card' do
+  .TeamMemberCard
+    .TeamMemberCard-image(style="background-image: url(#{user.avatar.url[:thumb]})")
```
scpaye506: this should be `user.avatar.url(:thumb)`, otherwise looks good

andylushman: 👍

PR #435, opened 2019-06-18
Get involved

andylushman:
* Updated `get-involve` page
* Added routes for `present` and `content` pages
* Created placeholder pages for `present` and `content`
* Changed the redirect for `/faq` from `get-involved/faq` to `about/faq`
* Commented out the test that were being used for the code on the previous `content` page.
* The naming is the buttons right now is not `learn more` but that will be address in my next PR that will allow for button naming to be passed in.

### DIFF
--- a/spec/features/content_pages_spec.rb
+++ b/spec/features/content_pages_spec.rb
@@ -1,84 +1,84 @@
-require "spec_helper"
+# require "spec_helper"
 
-feature "Content-only pages" do
-  scenario "the sponsors page" do
-    pending('refactor')
-    visit "/"
-    click_link "Sponsors"
-    expect(page).to have_content("OUR #{Date.today.year} SPONSORS")
-  end
+# feature "Content-only pages" do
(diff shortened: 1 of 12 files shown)


spec/features/content_pages_spec.rb
```diff
@@ -1,84 +1,84 @@
...
-feature "Content-only pages" do
-  scenario "the sponsors page" do
-    pending('refactor')
-    visit "/"
-    click_link "Sponsors"
-    expect(page).to have_content("OUR #{Date.today.year} SPONSORS")
-  end
+# feature "Content-only pages" do
+#   scenario "the sponsors page" do
+#     pending('refactor')
```
scpaye506: you shouldn't need to comment out any of these specs, instead just put `pending('refactor')` at the top of each test that is failing, like in this one.

andylushman: Nice! Just added it to the failing test 👍

PR #436, opened 2019-06-18
Content card button title

andylushman:
Now we can add custom `button_title` for the `content_cards` on the following pages...
*About
*Program
*Get Involved


app/views/site/about.html.slim
```diff
@@ -5,4 +5,4 @@ div.About-list-wrapper
   = render 'components/content_card_list', attributes: { 'aria-label': "About page" }  do
     - about_static_data[:items].each do |card|
       = render 'components/content_card_list_item' do
-        = render 'components/content_card', card: card, title: "#{card[:title]}", page_url: "about/#{card[:button_page_url]}"
+        = render 'components/content_card', card: card, button_title: "#{card[:button_title]}", title: "#{card[:title]}", page_url: "about/#{card[:button_page_url]}"
```
scpaye506: do we need to pass in the button_title as its own local here?

andylushman: good call! Just removed it from here and and now using it in `_content_card.html.slim`

app/views/site/program.html.slim
```diff
@@ -5,7 +5,7 @@ div.Program-list-wrapper
   = render 'components/content_card_list', attributes: { 'aria-label': "Programs" }  do
     - programs_static_data[:programs].each do |card|
       = render 'components/content_card_list_item' do
-        = render 'components/content_card', card: card, title: "Program", page_url: "program/"
+        = render 'components/content_card', card: card, button_title: "#{card[:button_title]}", title: "Program", page_url: "program/"
```
scpaye506: and here

andylushman: same here.

PR #439, opened 2019-06-19
Update profile avatar

scpaye506:
adds upload avatar to sign up and edit profile pages

needs a component from https://github.com/denverstartupweek/dsw-site/pull/429 to complete it though


scpaye506:
@andylushman you wont be able to upload it locally without some setup

PR #447, opened 2019-06-20
Faq page

andylushman:
Created the following
* Accordion list component
     * Accordion list item
     * `toggle_accordion.js` file
* Faq page now includes the accordion component which list the faq questions
* Faq page also has a hardcoded question about team members


app/views/site/about/faq.html.slim
```diff
@@ -1,4 +1,8 @@
...
+= render layout: 'layouts/shared/secondary_hero', locals: { title: 'FAQ', sub_title: 'Frequently asked questions', description: 'Denver Startup Week is a week of events showcasing Denver’s entrepreneur community in tech, design, business, and manufacturing.' } do
   = render 'layouts/shared/about_secondary_hero_nav'
 
-h1 Faq page
+.Faq-container
+  = render 'components/accordion_list', attributes: { 'aria-label': "Faq questions" }  do
+    - helpscout_articles_for_category('Website').each do |article|
+      = render 'components/accordion_list_item', title: process_with_liquid(article.name).html_safe, panel: process_with_liquid(article.content).html_safe
+    = render 'components/accordion_list_item', title: "Who is on the Denver Startup Week team?", panel: process_with_liquid("You can see a list of team members <a href='https://www.denverstartupweek.org/about/team'>here</a>").html_safe
```
scpaye506: where did this come from?

app/views/site/about/faq.html.slim
```diff
@@ -1,4 +1,8 @@
-= render layout: 'layouts/shared/secondary_hero', locals: { title: 'About', sub_title: 'By the community, for the community', description: 'Bringing Denver Startup Week to life takes the support, skills, and drive of an entire community: entrepreneurs, investors, civic leaders, educators, business pioneers, and more.' } do
+= render layout: 'layouts/shared/secondary_hero', locals: { title: 'FAQ', sub_title: 'Frequently asked questions', description: 'Denver Startup Week is a week of events showcasing Denver’s entrepreneur community in tech, design, business, and manufacturing.' } do
   = render 'layouts/shared/about_secondary_hero_nav'
 
-h1 Faq page
+.Faq-container
+  = render 'components/accordion_list', attributes: { 'aria-label': "Faq questions" }  do
```
scpaye506: i'm not seeing the hard coded faq that is stored in `layouts/shared/faq/contact

app/views/components/_accordion_list_item.html.slim
```diff
@@ -0,0 +1,7 @@
+li.AccordionListItem(data-bindable='toggle-accordion')
+  button.AccordionListItem-accordion
```
scpaye506: should we just put the bindable on the button here? then you don't have to do `.find(".AccordionListItem-accordion")` which is actually finding every button with that class, and it runs for every item. We probably don't want that.

andylushman: Good call! Just moved the event listener and changed `toggle_accordion`

app/views/site/about/faq.html.slim
scpaye506: but where did it come from? this is not the harded coded faq that's in `layouts/shared/faq/contact`

andylushman: I'll switch it so that the question matches the original one they had. I made this one up.

andylushman: Just changed it and created a partial.

app/views/site/about/faq.html.slim
andylushman: should be there now under _new_contact

PR #452, opened 2019-06-21
Content page

andylushman:
* Added `Content` page and the associated cards and links.
* Modified `content_card` so that it can accept the the new data and also accept links from `yml` file.
* Created a check to see the the user is signed-in and change the `Submit Article` button link based on signed-in in status.


app/data/content.yml
```diff
@@ -0,0 +1,43 @@
+items:
+  - title: 'Article Submissions'
+    description: 
+      - 'Lorem ipsum dolere sit amet nonummy consecuter adspecit nomen es quid. Lorem ipsum dolere sit amet nonummy consecuter  nomen es quid. Lorem ipsum dolere sit amet nonummy adspecit nomen es quid. Lorem ipsum dolere sit  nonummy consecuter adspecit nomen es quid. Lorem ipsum dolere sit amet nonummy consecuter adspecit es quid. Lorem ipsum dolere sit amet nonummy consecuter adspecit nomen es quid. Lorem ipsum dolere sit amet nonummy consecuter adspecit nomen es quid. Lorem ipsum dolere sit amet nonummy consecuter adspecit nomen es quid. Lorem ipsum dolere sit amet nonummy consecuter  nomen es quid. Lorem ipsum dolere sit amet nonummy adspecit nomen es quid. Lorem ipsum dolere sit  nonummy consecuter adspecit nomen es quid. Lorem ipsum dolere sit amet nonummy consecuter adspecit es quid. Lorem ipsum dolere sit amet nonummy consecuter adspecit nomen es quid. Lorem ipsum dolere sit amet nonummy consecuter adspecit nomen es quid.'
+    button_title: 'Submit Article'
+    button_page_url: 'users/sign_up'
```
scpaye506: this should be `articles/new`

app/views/site/get_involved/content.html.slim
```diff
@@ -1,7 +1,14 @@
-= render layout: 'layouts/shared/secondary_hero', locals: { title: 'Get Involved', sub_title: 'By the community, for the community', description: 'Bringing Denver Startup Week to life takes the support, skills, and drive of an entire community: entrepreneurs, investors, civic leaders, educators, business pioneers, and more.' } do
+= render layout: 'layouts/shared/secondary_hero', locals: { title: 'Content', sub_title: 'By the community, for the community', description: 'Bringing Denver Startup Week to life takes the support, skills, and drive of an entire community: entrepreneurs, investors, civic leaders, educators, business pioneers, and more.' } do
   = render 'layouts/shared/get_involved_secondary_hero_nav'
 
-h1 Content Page
+div.ContentPage-list-wrapper
+  = render 'components/content_card_list', attributes: { 'aria-label': "Content page" }  do
+    - content_static_data[:items].each do |card|
+      - if card[:button_page_url] == "users/sign_up" && signed_in?
```
scpaye506: this logic doesn't need to be here, it will auto - redirect them to sign in if they aren't already if they try to visit `articles/new`

app/views/components/_content_card.html.slim
```diff
@@ -10,11 +10,19 @@ div.ContentCard
     div.ContentCard-inner-content
       - if card[:description].kind_of?(Array)
         - card[:description].each do |desc|
-          p.ContentCard-description #{desc}
+          p.ContentCard-description = process_with_liquid(desc).html_safe
```
scpaye506: why are we using `process_with_liquid` in this file?

app/assets/stylesheets/redesign/views/content.sass
```diff
@@ -0,0 +1,21 @@
+.ContentPage-list-wrapper
+  margin: 62px auto 246px
+  padding: 0 5%
+  box-sizing: content-box
+  .ContentCard-list-item
```
scpaye506: so it looks like we have a new layout type on this page, but you are re-using the ContentCard by adding logic new html and logic to the content card component and then making it look right by adding the styles here.

This means if they ever have another page that wants to follow this same style they will have to replicate all of these styles again on that new page.

It seems like we probably need a new `content_block` type of component that encapsulates all of the new html/logic requirements and styles and just use that instead of re-working the ContentCard component.

Thoughts?

app/views/site/get_involved/content.html.slim
andylushman: Nice!

app/data/content.yml
andylushman: makes sense after reading your comment about being auto re-directed.

app/assets/stylesheets/redesign/views/content.sass
andylushman: Got it. I like that I idea. I will add a `content_block` component. 👍

app/views/components/_content_card.html.slim
andylushman: I was using this so that it would add the html tags from the yml file. For example, some of the content on the content page has links to other pages such as google forms. I can remove it from here and put it in the new `content_block` component.

scpaye506: `process_with_liquid` doesn't render the html the `.html_safe` method does that. I don't think you need `process_with_liquid`

andylushman: Got it. I will make the change. Thanks!

app/assets/stylesheets/redesign/views/content.sass
andylushman: just updated with `content_card_block`.

app/assets/stylesheets/redesign/components/content_card_block.sass
```diff
@@ -0,0 +1,74 @@
+@import ../config
+
+.ContentCardBlock
```
scpaye506: would you be opposed to just calling this `ContentBlock`? it's not really a card anymore

andylushman: Makes sense. just made the name change.

PR #453, opened 2019-06-21
Assets page

andylushman:
* Added assets page
* Uploaded new icon `download_assets`


app/views/site/about/assets.html.slim
```diff
@@ -1,4 +1,68 @@
-= render layout: 'layouts/shared/secondary_hero', locals: { title: 'About', sub_title: 'By the community, for the community', description: 'Bringing Denver Startup Week to life takes the support, skills, and drive of an entire community: entrepreneurs, investors, civic leaders, educators, business pioneers, and more.' } do
+= render layout: 'layouts/shared/secondary_hero', locals: { title: 'Assets', sub_title: 'Band Assets', description: 'Official Denver Startup Week graphical assets and styleguide are available for download. If you have more specific needs, please contact us.' } do
   = render 'layouts/shared/about_secondary_hero_nav'
 
-h1 Assets page
+.Assets
+  ul.Assets-list
+    li.Assets-list-item 
+      a.Assets-list-item-link href='/media_assets/DenverStartupWeek.eps'
```
scpaye506: it looks like we could move all of this content to a static data yml file and DRY up this html a bit

app/assets/stylesheets/redesign/views/assets.sass
```diff
@@ -0,0 +1,25 @@
...
+  
+.Assets
+  max-width: 1252px
+  margin: 84px auto 192px
+  padding: 0 5%
+  box-sizing: content-box
+
+.Assets-list-item
+  margin-bottom: 22px
+  color:
```
scpaye506: surprised this isn't throwing an error. but probably should remove it if not being used

app/views/icons/_download_assets.html.slim
```diff
@@ -0,0 +1,16 @@
+<svg width="15px" height="20px" viewBox="0 0 15 20" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: sketchtool 54.1 (76490) - https://sketchapp.com -->
+    <title>F0E50D70-4AF5-4866-840D-347EA75490C1@1.00x</title>
+    <desc>Created with sketchtool.</desc>
+    <g id="Internal-Views" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round">
+        <g id="FAQ-Open-Copy-2" transform="translate(-109.000000, -763.000000)" stroke="#05D3D8" stroke-width="1.5">
```
scpaye506: a pattern that we've been following with svgs is to change the stroke to `stroke=currentColor` and then the color property of the parent element will be used to color the svg. it makes it a bit more re-usable

app/views/site/about/assets.html.slim
andylushman: Just moved everything to a `yml` file.

app/data/assets.yml
```diff
@@ -0,0 +1,34 @@
...
+    url: "/media_assets/DENStartupWeek.eps"
+  - title: "DEN Startup Week Logo Reversed (EPS)"
+    url: "/media_assets/DENStartupWeek_Reversed.eps"
+  - title: "Chase Basecamp Logo (EPS)"
+    url: "/media_assets/ChaseBasecamp2017.pdf"
+  - title: "Geometric Mountains (EPS)"
+    url: "/media_assets/GEOMETRIC_MOUNTAINS.eps"
+  - title: "Style Guide (EPS)"
+    url: "/media_assets/DSW16_STYLE_GUIDE.pdf"
+    new_page: ""
```
scpaye506: instead of this `new_page` property, what do you think about a `target` property that can be `_blank` or `_self` and then it would be on each item here, making it a bit more consistent

you would also then be able to get rid of the if/else statement in `assets.html.slim`

andylushman: Nice! I like that! Done.

PR #460, opened 2019-06-24
added link to volunteer page

andylushman:
* Added link to volunteer page.

### DIFF
--- a/app/helpers/route_helper.rb
+++ b/app/helpers/route_helper.rb
@@ -22,7 +22,7 @@ module RouteHelper
   end
 
   def get_involved_routes
-    [get_involved_route, present_route, content_route]
+    [get_involved_route, present_route, volunteer_route, content_route]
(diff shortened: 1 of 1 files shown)


app/helpers/route_helper.rb
```diff
@@ -100,6 +100,13 @@ def get_involved_route
     }
   end
 
+  def volunteer_route
+    {
+      path: "https://www.cervistech.com/acts/console.php",
```
andylushman: @scpaye506 , this is the link that is provided in the story. However the link on the current site seems more correct: `https://www.cervistech.com/acts/console.php?console_id=0282&console_type=event&ht=1`

scpaye506: let's go with the current site url

andylushman: 👍

PR #463, opened 2019-06-25
Refactor authorship/publishing model

jayzes:
- Allow multiple authors on an article
- Move published_at to Publishing model
- Allow articles or submissions to be published

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -11,6 +11,7 @@ class Article < ApplicationRecord
   has_many :authors, class_name: "User", through: :authorships, source: :user
   has_one :publishing, as: :subject
 
+  belongs_to :submitter, class_name: "User"
   belongs_to :submission, required: false
   belongs_to :company, required: false
 
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -16,7 +16,8 @@ class Article < ApplicationRecord
 
   mount_uploader :header_image, HeaderImageUploader
 
-  accepts_nested_attributes_for :publishing
+  accepts_nested_attributes_for :publishing, allow_destroy: true
+  accepts_nested_attributes_for :authorships, allow_destroy: true
 
   def to_param
     "#{id}-#{title.try(:parameterize)}"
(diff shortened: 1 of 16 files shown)


app/models/article.rb
```diff
@@ -7,7 +7,12 @@ class Article < ApplicationRecord
     presence: true
 
   has_and_belongs_to_many :tracks, validate: false
-  belongs_to :author, class_name: "User"
+  has_many :authorships, dependent: :destroy
+  has_many :authors, class_name: "User", through: :authorships
+  has_one :publishing, as: :subject
+
+  belongs_to :submission, required: false
+  belongs_to :company, required: false
```
scpaye506: @jayzes were you going to also add a `submitter_id` to the article model?

jayzes: For sure!

db/migrate/20190624043458_create_authorships.rb
```diff
@@ -0,0 +1,20 @@
...
+  def up
+    create_table :authorships do |t|
+      t.references :article, foreign_key: true
+      t.references :user, foreign_key: true
+      t.boolean :is_displayed, default: true, null: false
+      t.timestamps
+    end
+
+    Article.find_each do |a|
+      a.update_attribute(:author_ids, [a.author.id])
```
scpaye506: looks like this needs to be `a.author_id` instead of `a.author.id`
since you removed `belongs_to :author, class_name: "User"` from the `article` model
otherwise it throws an error

db/migrate/20190624043458_create_authorships.rb
```diff
@@ -0,0 +1,20 @@
+class CreateAuthorships < ActiveRecord::Migration[5.2]
+  def up
+    create_table :authorships do |t|
+      t.references :article, foreign_key: true
+      t.references :user, foreign_key: true
```
scpaye506: should this be `t.references :author, foreign_key: {to_table: :users}, null: false` ? I'm getting some errors running the migration, with it not knowing how to save `author_ids` to articles

scpaye506: nevermind, ignore this.

we just need to update the article model to have this

`has_many :authors, class_name: "User", through: :authorships, source: :user`

was missing the `source: :user` part

PR #464, opened 2019-06-25
Content update

andylushman:
* changed styling so that it looks correct on `IE11`

### DIFF
--- a/app/assets/stylesheets/redesign/components/content_block.sass
+++ b/app/assets/stylesheets/redesign/components/content_block.sass
@@ -56,3 +56,9 @@
 @media(min-width: 1300px)
   .ContentBlock-content
     margin: 0 41px
+
+@media all and (-ms-high-contrast: none), (-ms-high-contrast: active)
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/redesign/components/content_block.sass
```diff
@@ -72,3 +56,9 @@
 @media(min-width: 1300px)
   .ContentBlock-content
     margin: 0 41px
+
+@media all and (-ms-high-contrast: none), (-ms-high-contrast: active)
+  /* IE10+ CSS styles go here */
+  @media(min-width: 640px)
+    .ContentBlock-action
+      display: flex
```
scpaye506: why do we need this? I thought flex was causing the issue.

andylushman: I think it might have something to do with the `mixin` that we are using for the button. Without it, the button stretches the full width of the screen in desktop.

PR #465, opened 2019-06-25
added styling for new footer

andylushman:
Footer update with `Stay In Touch` section

### DIFF
--- a/app/assets/stylesheets/redesign/views/footers.sass
+++ b/app/assets/stylesheets/redesign/views/footers.sass
@@ -44,10 +44,7 @@
   text-transform: uppercase
   margin-bottom: 31px
   
-.Footer-updates-email
-  +font-family-mixin(Montserrat-Light)
-  font-size: 14px
-  border-bottom: 1px solid $primary-font-light
+
 
 .Footer-updates-submit
   +white-button(166px, 46px)
--- a/app/assets/stylesheets/redesign/views/footers.sass
+++ b/app/assets/stylesheets/redesign/views/footers.sass
@@ -1,6 +1,8 @@
 @import ../config
 
 .Footer
+  .SocialMediaNav
+    margin: -0.5em
   .Footer-title
     display: inline-block
     vertical-align: middle
@@ -23,30 +25,13 @@
 .Footer-block
   &:first-child
     margin-bottom: 50px
-  a
-    text-decoration: underline
-    color: #4d4d4d
-  nav
-    margin: -0.5em
-
-.Footer-page-links-list
-  margin-top: 45px
-
-.Footer-link
-  margin-top: 1em
-  font-weight: 500
-  font-size: 12px
-
-.Footer-link a
-  text-decoration: none
-  text-transform: uppercase
 
 .Footer
   .SocialMediaNav-fa-circle
     background: #4d4d4d
   .SocialMediaNav-icon
     color: #fff
-
+ 
 .Footer-primary-callouts
   margin-top: 80px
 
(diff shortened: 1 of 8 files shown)


app/views/layouts/shared/_footer.html.slim
```diff
@@ -4,16 +4,20 @@ footer.Footer
...
 
     .Footer-block
-      div.Footer-title Denver Startup Week #{Date.today.year}
-      div.Footer-subtitle &copy;Copyright #{Date.today.year}. All rights reserved.
-      = render 'components/social_media_nav'
\ No newline at end of file
+      .Footer-updates 
+        p.Footer-updates-title Stay in Touch
+        = form_for NewsletterSignup.new, class: "footer-updates-form", html: { data: { bindable: "newsletter-signup" } } do |f|
+          = f.text_field :email, placeholder: "Email Address", class: "Footer-updates-email"
```
scpaye506: you could use the `floating_input` component here. 

```
= render layout: 'components/floating_input' do
  = f.text_field ....
  = f.label ...
```
label has to go after the text_field with how it's currently setup

app/views/layouts/shared/_footer.html.slim
```diff
@@ -4,16 +4,20 @@ footer.Footer
...
+      .Footer-title Denver Startup Week #{Date.today.year}
+      .Footer-subtitle &copy;Copyright #{Date.today.year}. All rights reserved.
+      = render 'components/social_media_nav'
+      ul.Footer-page-links-list aria-label="Footer page links"
         - if signed_in? && current_user.is_admin?
           li.Footer-link= link_to 'Admin', admin_root_path
-        li.Footer-link= link_to 'Attendees', page_path(page: 'attendees')
-        li.Footer-link= link_to 'Podcast', page_path(page: 'podcast')
-        li.Footer-link= link_to 'Press', page_path(page: 'press')
+        li.Footer-link= link_to 'FAQ', page_path(page: 'about/faq')
```
scpaye506: thoughts on putting the footer routes in the `route_helper`?

app/assets/stylesheets/redesign/views/footers.sass
```diff
@@ -44,6 +50,31 @@
...
+  margin-bottom: 31px
+  
+.Footer-updates-email
+  +font-family-mixin(Montserrat-Light)
+  font-size: 14px
+  border-bottom: 1px solid $primary-font-light
+
+.Footer-updates-submit
+  +white-button(166px, 46px)
+  width: fit-content
```
scpaye506: i'm not really sure we should be using `fit-content` it's not actually part of the spec and not supported by all browsers
https://developer.mozilla.org/en-US/docs/Web/CSS/width

If the `white-button` mixin is causing problems we should probably refactor that to work instead or just sit a fixed width where needed

app/views/layouts/shared/_footer.html.slim
andylushman: Nice!

app/views/layouts/shared/_footer_nav.html.slim
```diff
@@ -0,0 +1,7 @@
+nav.FooterNav(aria-label="#{nav_title} nav")
+  ul.FooterNav-page-links-list
+    - if signed_in? && current_user.is_admin?
```
scpaye506: I think we might want to move this logic to the `router_helper`, otherwise looks good to go

andylushman: 👍 Just moved it over.

PR #470, opened 2019-06-26
added sponsor page

andylushman:
Added `sponsor` link and sponsor page where user can download sponsorship prospectus


scpaye506:
this seems to be missing all of the current sponsors logic and data

andylushman:
> this seems to be missing all of the current sponsors logic and data

Carl mentioned that this `Sponsor` page is only for users who are interested in becoming a sponsor and thus only needs to be able to download the sponsorship prospectus. Another page `sponsors`, will list all of the sponsors.

app/data/sponsor.yml
```diff
@@ -0,0 +1,6 @@
+items:
+  - title: 'Become a Sponsor'
+    description: 
+      - 'Denver Startup Week is wholly financed through sponsorships that leverage the resources and capacities of the entire community – ultimately ensuring that Denver Startup Week exists for the community, by the community.'
+    button_title: 'Download Sponsorship Prospectus'
+    button_page_url: 'https://www.denverstartupweek.org/DSW%20Sponsorship%202019.pdf'
```
scpaye506: I think this should just be `'/DSW%20Sponsorship%202019.pdf'`

scpaye506:
> > this seems to be missing all of the current sponsors logic and data
> 
> Carl mentioned that this `Sponsor` page is only for users who are interested in becoming a sponsor and thus only needs to be able to download the sponsorship prospectus. Another page `Sponsors`, will list all of the sponsors.

Ok, that makes sense!

app/data/sponsor.yml
andylushman: Yeah, I thought the same thing and then realize that the `content_block` component is setup with `- if card[:button_title].present? && card[:button_page_url].include?("https://") || card[:button_page_url].include?("mailto:") `. Just adding `'/DSW%20Sponsorship%202019.pdf'` won't open it in a new tab. Should I change the logic in the `content_block` or use the full url?

scpaye506: yeah let's update the logic in the `content_block`, maybe add `target` to the yml files

andylushman: Sounds good! Everything seems to be working fine except the `get-involved/articles/new` route. Do you have any ideas about why that might be down?

scpaye506: the route is just `/articles/new`

andylushman: 👍  Should be good now.

PR #475, opened 2019-06-27
Flash messages

scpaye506:


app/javascript/components/flash_message.js
```diff
@@ -0,0 +1,33 @@
+import Bindable from 'utensils/bindable';
+
+export default class FlashMessage {
+  constructor(el) {
+    this.el = el;
+
+    this.closeFlashMessage = this.closeFlashMessage.bind(this);
```
andylushman: should we make these functions arrow functions so we wouldn't have to bind this way?

scpaye506: i think it's probably ok to use bind this way, and  it's also the convention they use in their other javascript files. If we were re-writing all of the javascript though I'd agree better to use arrow functions.

andylushman: 👍

PR #478, opened 2019-06-27
Common multi checkboxes

scpaye506:


app/controllers/registrations_controller.rb
```diff
@@ -43,7 +43,7 @@ def registration_params
                                            :learn_more_pledge_1p,
                                            :track_id,
                                            :coc_acknowledgement,
-                                           :attendee_goal_ids)
+                                           attendee_goal_ids: [])
```
scpaye506: @jayzes noticed when i was making this change that `attendee_goal_ids` param was not being permitted since it's an array. hopefully this will get you more attendee goals next registration period!

jayzes: Ah crap, that'd do it. Good catch! That's what I get for not adding tests around that feature.

PR #479, opened 2019-06-27
Linkedin link

andylushman:
Added `user_linkedin` component and made some styling adjustments so that it looked correct on desktop and also works for mobile.


app/views/schedules/show.html.slim
```diff
@@ -76,6 +76,8 @@
           - @session.registrants.where.not(id: @current_user.id).reorder('"session_registrations"."created_at" DESC').each do |user|
             .SessionDetail-attendee
               = render partial: 'components/name_with_avatar', locals: { user: user, align: 'right', color: 'gray' }
+              - if user[:linkedin_url].present?
```
scpaye506: we should probably add this to the signed in user, a few lines above, as well

andylushman: Sounds good!

PR #481, opened 2019-06-28
Added titles to pages

andylushman:
Added titles to pages. We will have to add titles to any pages that we create from now onward. For example, I will add a page tittle to the `sponsors` page when I create it.

### DIFF
--- a/app/views/articles/_article_form.html.slim
+++ b/app/views/articles/_article_form.html.slim
@@ -1,5 +1,4 @@
 - content_for(:fullscreen_takeover, true)
-- content_for(:title, 'Submit Article')
 - upload_label = @article.header_image.present? ? 'Change Header Image +' : 'Add Header Image +'
 
 = render 'layouts/shared/fullscreen_takeover_header', title: content_for(:article_page_title), background: 'light'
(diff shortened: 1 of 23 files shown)


app/views/site/program/tracks/show.html.slim
```diff
@@ -1,2 +1,4 @@
+- content_for(:title, 'Programs')
```
scpaye506: Should this title be `Tracks`?

app/views/articles/_article_form.html.slim
```diff
@@ -1,4 +1,5 @@
 - content_for(:fullscreen_takeover, true)
+- content_for(:title, 'Submit Article')
```
scpaye506: this is a partial, we should move the title to the `articles/new` and `articles/edit` pages. and `articles/show` probably needs something as well

app/views/site/program/tracks/show.html.slim
andylushman: Good catch!

app/views/articles/_article_form.html.slim
andylushman: good call. made titles for `Submit article`, `Edit Article` and on the `articles/show`...
```
- content_for(:title, 'Articles')
- content_for(:subtitle, "#{@article[:title]}")
```

PR #482, opened 2019-06-28
updated feedback_form

andylushman:
Updated `feedback_form`.

### DIFF
--- a/app/views/schedules/_feedback_form.html.slim
+++ b/app/views/schedules/_feedback_form.html.slim
@@ -6,6 +6,6 @@ section.SessionFeedback
       = f.label :rating, 'Please rate this session'
     = render layout: 'components/floating_input' do
       = f.text_field :comments
-      = f.hidden_field :submission_id, value: @session.id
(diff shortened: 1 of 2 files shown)


app/views/schedules/_feedback_form.html.slim
```diff
@@ -1,13 +1,11 @@
...
+    = render layout: 'components/floating_input' do
+      = f.select :rating, ratings_for_select, { :include_blank => '' }, required: 'required'
       = f.label :rating, 'Please rate this session'
-      = f.select :rating, ratings_for_select, required: 'required'
-
-      = f.label :comments, 'Comments'
+    = render layout: 'components/floating_input' do
       = f.text_field :comments
-
       = f.hidden_field :submission_id, value: @session.id
```
scpaye506: this hidden field probably shouldn't be nested under the `floating_input`, otherwise looks good

andylushman: Just moved it 👍

PR #483, opened 2019-07-01
Publishings filter and search

scpaye506:
adds ability to filter publishings for articles and submissions

Might not be the most performant - I'm searching articles and submissions individually and getting the id's and then passing those subject id's to the Publishing model. I don't think there is a performant elegant solution though with the polymorphic relationship

Added some logic for when Submissions are shown based on the annual schedule, not sure if that logic makes sense or not though.

### DIFF
--- a/app/models/publishing.rb
+++ b/app/models/publishing.rb
@@ -3,8 +3,12 @@ class Publishing < ApplicationRecord
 
   def self.filtered_results(filters)
     articles = Article.for_publishings_filter(filters)
-    sessions = AnnualSchedule.registration_open? || AnnualSchedule.post_week? ? Submission.for_publishings_filter(filters) : []
+    sessions = []
+    if AnnualSchedule.registration_open? || AnnualSchedule.post_week?
+      sessions = Submission.for_publishings_filter(filters)
+    end
     ids = articles.pluck(:id) + sessions.pluck(:id)
+
     where(subject_id: ids)
     .order(effective_at: :desc)
   end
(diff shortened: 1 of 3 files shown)


app/models/publishing.rb
```diff
@@ -1,3 +1,13 @@
 class Publishing < ApplicationRecord
   belongs_to :subject, polymorphic: true
+
+  def self.filtered_results(filters)
+    articles = Article.for_publishings_filter(filters)
+    sessions = AnnualSchedule.registration_open? || AnnualSchedule.post_week? ?
```
scpaye506: @jayzes does this make sense to only include submissions in the publishings stream when registration is open or during post week?

app/models/submission.rb
```diff
@@ -103,21 +103,21 @@ def self.for_submittable_tracks
...
-    else
-      all
-    end
+    return all if name == 'all' || name.blank?
+    joins(:cluster).where("LOWER(clusters.name) = LOWER(?)", name)
+  end
+
+  def self.for_publishings_filter(filters)
+    return all if filters.blank?
+    .where(year: Date.today.year)
```
scpaye506: @jayzes does this make sense for submissions to always be filtered by the current year?

jayzes:
Also, agreed on the lack of eager loading for polymorphic joins - that is one of the drawbacks there for sure. I'll take a look and see if anything else there jumps out at me, but probably fine for now.

PR #484, opened 2019-07-01
download sponsor prospectus wasn't nested under if statement

andylushman:
noticed that the option to download the sponsor prospectus wasn't nested under the `if` statement.

### DIFF
--- a/app/views/site/get_involved/sponsor.html.slim
+++ b/app/views/site/get_involved/sponsor.html.slim
@@ -4,8 +4,8 @@
   = render 'layouts/shared/get_involved_secondary_hero_nav'
 
 - if AnnualSchedule.sponsorship_open?
-section.Sponsor-wrapper
-  = render 'components/content_block_list', attributes: { 'aria-label': "Content page" }  do
-    - sponsor_static_data[:items].each do |sponsor|
-      = render 'components/content_block_list_item' do
-        = render 'components/content_block', card: sponsor, title: "#{sponsor[:title]}", page_url: "#{sponsor[:button_page_url]}"
+  section.Sponsor-wrapper
+    = render 'components/content_block_list', attributes: { 'aria-label': "Content page" }  do
+      - sponsor_static_data[:items].each do |sponsor|
+        = render 'components/content_block_list_item' do
+          = render 'components/content_block', card: sponsor, title: "#{sponsor[:title]}", page_url: "#{sponsor[:button_page_url]}"


scpaye506:
Since the sole purpose of this page is to show the sponsorship prospectus, when sponsorship is not open, should we hide the entire route or show some sort of not open message?

andylushman:
> Since the sole purpose of this page is to show the sponsorship prospectus, when sponsorship is not open, should we hide the entire route or show some sort of not open message?

Good question. At first, I thought it would be more clean to remove the link but after thinking about it. If a user comes to the site with the intention of becoming a sponsor and doesn't see a link, they will have to reach out to someone or they may even just leave the site and not return. However, if we leave them a message, they will at least be informed. Thoughts?

scpaye506:
> > Since the sole purpose of this page is to show the sponsorship prospectus, when sponsorship is not open, should we hide the entire route or show some sort of not open message?
> 
> Good question. At first, I thought it would be more clean to remove the link but after thinking about it. If a user comes to the site with the intention of becoming a sponsor and doesn't see a link, they will have to reach out to someone or they may even just leave the site and not return. However, if we leave them a message, they will at least be informed. Thoughts?

I'd have Carl check with Rye

andylushman:
> > > Since the sole purpose of this page is to show the sponsorship prospectus, when sponsorship is not open, should we hide the entire route or show some sort of not open message?
> > 
> > 
> > Good question. At first, I thought it would be more clean to remove the link but after thinking about it. If a user comes to the site with the intention of becoming a sponsor and doesn't see a link, they will have to reach out to someone or they may even just leave the site and not return. However, if we leave them a message, they will at least be informed. Thoughts?
> 
> I'd have Carl check with Rye

Sounds good. I'll reach out to Carl.

andylushman:
> > > > Since the sole purpose of this page is to show the sponsorship prospectus, when sponsorship is not open, should we hide the entire route or show some sort of not open message?
> > > 
> > > 
> > > Good question. At first, I thought it would be more clean to remove the link but after thinking about it. If a user comes to the site with the intention of becoming a sponsor and doesn't see a link, they will have to reach out to someone or they may even just leave the site and not return. However, if we leave them a message, they will at least be informed. Thoughts?
> > 
> > 
> > I'd have Carl check with Rye
> 
> Sounds good. I'll reach out to Carl.

@scpaye506 , from Rye via Carl
```
"We have never had an on/off, so we have generally just left the download prospectus up and then I have left folks know via email that it is closed. That is a fine approach for now, unless there is an obvious way to make the prospectus download box easily disappear."
```
Carl said that he would rather leave it up year-round.

PR #490, opened 2019-07-02
Code of conduct update

andylushman:
Locally, closing the fullscreen take-over works fine on the `code-of-conduct` page. However, in the staging env, when clicking the close button the user stays on the `code-of-conduct` page. In this PR I am passing `on_close` directly from the `code-of-conduct` page to see if that helps in the staging env.

### DIFF
--- a/app/views/site/code_of_conduct.html.slim
+++ b/app/views/site/code_of_conduct.html.slim
@@ -1,6 +1,7 @@
 -content_for(:fullscreen_takeover, true)
+- on_close = (request.referer =~ /(code-of-conduct)/i).nil? ? request.referer : dashboard_path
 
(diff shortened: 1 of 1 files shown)


app/views/site/code_of_conduct.html.slim
```diff
@@ -1,7 +1,8 @@
 -content_for(:fullscreen_takeover, true)
 - content_for(:subtitle, 'Code of Conduct')
+- on_close = (request.referer =~ /(code-of-conduct)/i).nil? ? request.referer : dashboard_path
```
scpaye506: i think this can just be `request.referer || '/'`

andylushman: Makes sense!

PR #499, opened 2019-07-06
 Allow users to opt out of public attendee lists

jayzes:
- Get profile editing feature spec working again
- Simplify public attendee selection/ordering logic and add tests
- Take a first stab at UI (though it needs styling help)

This will throw a (misleading) deprecation warning until rails/rails#35993 is fixed.


jayzes:
Interestingly, this seems not to fail in isolation (locally) but it does when run as part of the suite.

scpaye506:
> Interestingly, this seems not to fail in isolation (locally) but it does when run as part of the suite.

oh that's strange. I just saw that it did just pass with the way you have it. Any idea what was going on there?

jayzes:
I'm not sure - it seems to be working now that it's merged in, so I'm going to chalk that one up to a spurious failure unless it starts cropping up repeatedly.

PR #500, opened 2019-07-07
Refactor article search query and add FT indices

jayzes:


scpaye506:
@jayzes is there a migration file that didn't get committed?

jayzes:
@scpaye506 you are correct! just fixed and re-pushed.

PR #507, opened 2019-07-08
Articles bug fixes

scpaye506:


app/models/publishing.rb
```diff
@@ -2,10 +2,10 @@ class Publishing < ApplicationRecord
   belongs_to :subject, polymorphic: true
 
   def self.filtered_results(filters)
-    article_ids = Article.for_publishings_filter(filters).select(:id)
-    session_ids = Submission.for_publishings_filter(filters).select(:id)
-    where(subject_type: "Article", id: article_ids)
-      .or(where(subject_type: "Submission", id: session_ids))
+    article_ids = Article.for_publishings_filter(filters).to_a.pluck(:id)
```
scpaye506: @jayzes i'm getting a `ERROR: subquery has too many columns` and it seems like it's because the `textacular` gem results from a `fulltext_search` do not really respect the ActiveRecord model anymore, so things like `.count` and `.select(:id)` don't work. In this instance `.select(:id)` just returns the full ActiveRecord result, not just the results with only the id column selected. Which is causing the subquery has too many columns error since we are passing an object with more than the id column in the where statement further down.

The easiest approach i've found is to just turn it into an array and then pluck out the id column. Not sure exactly of the performance consequences.

Any other approach you'd like to take with this?

jayzes: Ah, hmm. That's frustrating. Let me poke at it real quick and see if anything else jumps out at me.

scpaye506: @jayzes i'm going to merge this for now as it's breaking some functionality on the site and for the demo today. I created an issue for it and referenced this PR and assigned it to you for any future digging you want to do on it.

PR #512, opened 2019-07-08
Sponsors height

andylushman:
added min-height for sponsors `p` tag for desktop. This forces sponsors of the same type to be the same height, regardless of description length. In mobile, each sponsors card is still responsive based on content in description up until the line limit. 

Added line limit as per Bryan's comment and according to the mocks.

### DIFF
--- a/app/views/components/_sponsors_card.html.slim
+++ b/app/views/components/_sponsors_card.html.slim
@@ -1,4 +1,4 @@
-- header_image = sponsorship.logo.url || ""
+- header_image = sponsorship.logo.url || "https://cdn-images-1.medium.com/max/800/1*amnvhplMQHkIrq_iYD8GUA.jpeg"
 
 .SponsorsCard
   div.(class="SponsorsCard-image #{level}" style="background-image: url(#{image_url header_image})")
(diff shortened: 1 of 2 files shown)


andylushman:
Was going off of Bryan's comment in Slack but I will ask Carl to reach out to Rye to confirm.
image

andylushman:
@scpaye506 , found the issue with the sponsors not being the same size. There was still an old class name being used. Everything should be working fine now, with regards to the buttons lining up at the bottom. Only thing left is adding the character limits.

app/views/components/_sponsors_card.html.slim
```diff
@@ -14,5 +14,5 @@
           a(href="#{sponsorship[:link_href]}" aria-label="Visit Website" target="_blank" class="SponsorsCard-link")
             span.SponsorsCard-link-title Visit Website
             = render partial: 'icons/arrow_right'
-  - if level == "partner" || level == "member"    
+  - if level == "partner" || level == "member" || level == "field-guide"
```
scpaye506: it actually seems like there are a lot of other level types, should we just make the last if statement an if else and let it pick up everything that isn't a 6 card, or 5 card level?

andylushman: Good idea 👍

PR #515, opened 2019-07-09
Also interested in articles

scpaye506:


app/models/article.rb
```diff
@@ -62,7 +62,7 @@ def self.published
   end
 
   def related
-    base_query = self.class.left_outer_joins(:tracks, :authors)
+    base_query = self.class.includes(:tracks, :authors)
```
scpaye506: @jayzes i changed this to `includes` instead of `left_outer_joins`, I think it still accomplishes the same thing, but the join was meaning we could get duplicate records (1 for each associate track and for each associated author).
Another approach could be adding a distinct operator i suppose. Do you have another approach or preference for this?

jayzes: The reason I did it as a left outer join initially is that the regular join/include would only pull back results where there were all of the associations set (i.e. track and authors and company), where I feel like would want to match disjoint on any of those. Distinct might be better for that reason.

scpaye506: makes sense. updated to use distinct instead.

PR #521, opened 2019-07-09
Article admin body options and styles

scpaye506:


jayzes:
One thing I just noticed - is there any reason we want to keep subscript/superscript? I think it was in the original stock config that I dropped in but I don't know if it's actually needed for formatting articles.

scpaye506:
> One thing I just noticed - is there any reason we want to keep subscript/superscript? I think it was in the original stock config that I dropped in but I don't know if it's actually needed for formatting articles.

@jayzes I'm guessing it will almost never get used, I think the more simple the better. Want me to take it out?

jayzes:
Let's do it! I'm all for simplifying wherever possible.

PR #524, opened 2019-07-09
added open graph meta tags

andylushman:
added open graph meta tags for article details page.

### DIFF
--- a/app/views/articles/show.html.slim
+++ b/app/views/articles/show.html.slim
@@ -1,6 +1,12 @@
 - content_for(:title, 'Articles')
 - content_for(:subtitle, "#{@article[:title]}")
 - related_articles = @article.related.limit(3)
+- content_for(:extra_meta_tags) do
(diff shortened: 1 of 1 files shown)


andylushman:
> nice. I think we want the image as well. using the thumb image

Just added `meta(property="og:image" content="#{@article.header_image.url(:article_page)}")`

app/views/articles/show.html.slim
```diff
@@ -1,6 +1,13 @@
 - content_for(:title, 'Articles')
 - content_for(:subtitle, "#{@article[:title]}")
 - related_articles = @article.related.limit(3)
+- content_for(:extra_meta_tags) do
+  meta(property="og:title" content="Denver Startup Week | #{@article.title}")
+  meta(property="og:image" content="#{@article.header_image.url(:article_page)}")
```
scpaye506: use `:article_card` here instead of `:article_page` and then good to go

andylushman: 👍

PR #529, opened 2019-07-10
Copy update

andylushman:
optimized and updated images along with copy for various pages.


app/views/components/_content_card.html.slim
```diff
@@ -15,4 +15,7 @@ div.ContentCard
       - else
         p.ContentCard-description #{card[:description]}
       .ContentCard-action
-        = link_to button_title, page_path(page: page_url), class: 'ContentCard-link'
+        - if card[:target].present?
+          = link_to "#{card[:button_title]}", "#{page_url}", :target => "#{card[:target]}", class: 'ContentCard-link'
+        - else
+          = link_to button_title, page_path(page: page_url), class: 'ContentCard-link'
```
scpaye506: should these `link_to`'s be different? one uses `card[:button_title]` the other uses `button_title`, one uses `page_path`

andylushman: you're right. Just changed it to `button_title`.

PR #530, opened 2019-07-10
Safari radio

andylushman:
created custom radio button and change the code to implement it.


app/assets/stylesheets/redesign/components/radio_buttons.sass
```diff
@@ -0,0 +1,45 @@
+@import ../config
+  
+// CUSTOM RADIO BUTTON
+[type="radio"]
```
scpaye506: i think it might be worth putting all of this ins the `form_field.sass` file nested under the `.FormField` class, that way only radio buttons under the `form_field` component will be styled like this. Incase they need other radio styles else where.

andylushman: good call! Just moved everything over to `form_field.sass` and put it under the `.FormField` class.

PR #533, opened 2019-07-10
primary calllout nav now displays in block in windows

andylushman:
in `IE11` the `height` of the `flex` container was collapsing so I changed it to `display: block` and it now looks fine. One note is that on `IE11` each `li` in the `primary_callouts_nav.html.slim` can now have a different height, since it is display `block` instead of `flex`. Meaning, the arrow buttons may not lineup at the bottom.

### DIFF
--- a/app/assets/stylesheets/redesign/layouts/primary_callouts_nav.sass
+++ b/app/assets/stylesheets/redesign/layouts/primary_callouts_nav.sass
@@ -47,3 +47,8 @@
 
     &:last-child
       border-bottom: 0
+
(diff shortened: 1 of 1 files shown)


andylushman:
> pretty sure there is a flex solution to this. have you tried using `flex-grow: 1` on the list desc class?

Yeah, I tried that but it didn't work. I can keep looking for a flex solution.

andylushman:
> > pretty sure there is a flex solution to this. have you tried using `flex-grow: 1` on the list desc class?
> 
> Yeah, I tried that but it didn't work. I can keep looking for a flex solution.

looks like it needed `flex-basis: auto`

PR #535, opened 2019-07-11
Add featured_on_homepage to publishings model

### LINKED ISSUE #517: Add logic to populate "featured sessions & articles" on the homepage
Probably as an attribute on the publishings model

jayzes:
Allow it to be edited in the admin as well

Fixes #517

### DIFF
--- a/app/admin/articles.rb
+++ b/app/admin/articles.rb
@@ -9,7 +9,7 @@ ActiveAdmin.register Article do
     :company_id,
     :submission_id,
     :video_url,
-    publishing_attributes: [:id, :_destroy, :effective_at],
+    publishing_attributes: [:id, :_destroy, :effective_at, :featured_on_homepage],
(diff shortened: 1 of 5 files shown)


jayzes:
@ryefinegan @scpaye506 i didn't put in the randomization logic (yet) but figure this can get us started

scpaye506:
awesome! Should this also be added to Submissions?

jayzes:
@scpaye506 just squashed this down to a single commit - I'm guessing I should probably cherry-pick it back to the `staging` branch? as well? I don't think #537 needs the same treatment though I could be wrong.

scpaye506:
@jayzes I don't think any need to cherry pick to staging. I'll just merge master directly into staging once this and https://github.com/denverstartupweek/dsw-site/pull/537 are both merged

PR #537, opened 2019-07-11
Move publishing/unpublishing actions to buttons

jayzes:


scpaye506:
nice! Should you also add this for Submissions?

PR #554, opened 2019-07-12
Profile form updates

scpaye506:


app/assets/stylesheets/redesign/components/floating_input.sass
```diff
@@ -15,9 +15,15 @@
   transition: all 200ms
   opacity: 0.8
   pointer-events: none
+  width: 100%
+  white-space: nowrap
+  overflow: hidden
+  text-overflow: ellipsis
 
 .FloatingInput
+    input:focus ~ label,
```
andylushman: Looks like this extra indentation might be causing some test to fail.

PR #567, opened 2019-07-13
Add notification email for new articles

jayzes:


jayzes:
ENV var is set in all environments as well

PR #568, opened 2019-07-14
Add initial specs around submitting articles

### LINKED ISSUE #566: Add tests around article creation (for end users)
Could be part of #518 but worth calling out separately since this is net-new functionality

jayzes:
Fixes #566


jayzes:
Future @scpaye506 - I wasn't 100% sure how to get to the edit page for an article from the user perspective (though I saw the controller action for it). Is that currently linked up?

Issue #513, opened 2019-07-09
Publishings textacular issues


scpaye506:
@jayzes i'm getting a `ERROR: subquery has too many columns` and it seems like it's because the `textacular` gem results from a `fulltext_search` do not really respect the ActiveRecord model anymore, so things like `.count` and `.select(:id)` don't work. In this instance `.select(:id)` just returns the full ActiveRecord result, not just the results with only the id column selected. Which is causing the subquery has too many columns error since we are passing an object with more than the id column in the where statement further down.

The easiest approach i've found is to just turn it into an array and then pluck out the id column. Not sure exactly of the performance consequences.

Any other approach you'd like to take with this?

_Originally posted by @scpaye506 in https://github.com/denverstartupweek/dsw-site/pull/507_

jayzes:
One other thing I just noticed on this front - the `/articles` page on preview is taking a long time to load because the trivial case where no criteria are specified results in all submission IDs being loaded in. Ging to look at that too.

scpaye506:
> One other thing I just noticed on this front - the `/articles` page on preview is taking a long time to load because the trivial case where no criteria are specified results in all submission IDs being loaded in. Ging to look at that too.

@jayzes adding `where.not(publishing: nil)` should help with that quite a bit. I'll do a PR for it

PR #569, opened 2019-07-14
Fix up Textacular query for publishings

jayzes:
- The subquery column error was coming up because of this line in Textacular where it explicitly sets select/order: https://github.com/textacular/textacular/blob/master/lib/textacular.rb#L127, also referenced in https://github.com/textacular/textacular/issues/86. This commit reorders/excludes those criteria to allow the Textacular query to just return an ID, which is what the `IN` requires.
- Also fix a bug where posts without authors were not findable due to the inner join
- Add tests too!

Fixes https://github.com/denverstartupweek/dsw-site/issues/513

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -56,7 +56,7 @@ class Article < ApplicationRecord
       body: terms,
       users: {name: terms},
     }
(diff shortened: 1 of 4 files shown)


scpaye506:
nice!

PR #572, opened 2019-07-15
Add ability for users to be able to edit their articles

scpaye506:


app/models/article.rb
```diff
@@ -72,4 +72,8 @@ def related
 
     query.published.where.not(id: id).distinct
   end
+
+  def editable?
+    false
```
scpaye506: @jayzes want me to just leave this false for now until the editable logic is added? this returning true will show the edit icon on editable article cards in the dashboard

jayzes: That's perfect!

PR #574, opened 2019-07-15
Move tracks back to program for now

scpaye506:
@jayzes moving tracks back to the site/program dir for now. Not sure if you moved this on purpose but it breaks some functionality without changing logic in a few other places. Moving it back for now to resolve the current errors. If you want it moved back to the views root then let me know.


jayzes:
Sorry about that! I was too optimistic in my refactor. Definitely no need for it to be at the top level.

Issue #578, opened 2019-07-15
Staging S3 bucket permissions


scpaye506:
looks like track images uploaded in staging are getting 403 errors when trying to view them

can see the error in the console on this page

https://denver-startup-week-stage.herokuapp.com/program/tracks

```
tracks:2 GET https://dsw-staging-uploads.s3.amazonaws.com/uploads/track/header_image/content_card_CamNewton.jpg 403 (Forbidden)
```

Looks like it might be happening on sometimes on the preview site as well

jayzes:
Checking this out now

jayzes:
Weirdly, somehow the ID is getting lost off of the image URL:

- The request is going to https://dsw-staging-uploads.s3.amazonaws.com/uploads/track/header_image/content_card_CamNewton.jpg
- The actual image lives at https://dsw-staging-uploads.s3.amazonaws.com/uploads/track/header_image/3/content_card_CamNewton.jpg

Looking into reasons why that might be happening.

PR #584, opened 2019-07-15
Correct image URLs for track page headers

jayzes:
Explicitly format the model instances as a hash

Fixes https://github.com/denverstartupweek/dsw-site/issues/578

### DIFF
--- a/app/models/cluster.rb
+++ b/app/models/cluster.rb
@@ -13,9 +13,14 @@ class Cluster < ApplicationRecord
   end
 
   def self.template_list_data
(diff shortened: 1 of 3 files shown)


jayzes:
Did some quick checking locally and I *think* this should be backwards-compatible with the data file originated content, but worth double-checking.

scpaye506:
> Did some quick checking locally and I _think_ this should be backwards-compatible with the data file originated content, but worth double-checking.

looks like it loads the data file content just fine

PR #587, opened 2019-07-15
Some article updates

scpaye506:


app/views/articles/show.html.slim
```diff
@@ -22,7 +23,7 @@
       - @article.tracks.each do |track|
         =render partial: 'components/track_indicator', locals: { size: 'small', color: "#{track[:color]}", name: "#{track[:name]}" }
     - if @article.publishing.present?
-      .Article-date #{@article.publishing.effective_at.strftime("%A, %b%e, %Y")}
+      .Article-date #{@article.publishing.effective_at.strftime("%A, %b %e, %Y")}
```
andylushman: should we have  a `,` after `%b`?

scpaye506: no that would make the date `Monday, Jul, 15, 2019`. This is adding a space between Jul and 15

andylushman: got it!

PR #590, opened 2019-07-15
remvoed margin padding and max-width from footer and added it back to…

andylushman:
removed max-width and padding from `footer` and therefore primary-callouts. Added it back to `footer-blocks` and `footer-sponsors`

### DIFF
--- a/app/assets/stylesheets/redesign/views/footers.sass
+++ b/app/assets/stylesheets/redesign/views/footers.sass
@@ -100,5 +100,5 @@
     min-width: 350px
 
 @media(min-width: 800px)
-  .Footer
+  .Footer-blocks
     padding: 60px 20px 70px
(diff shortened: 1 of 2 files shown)


andylushman:
> This still looks a bit off from the design. I think the entire footer can be wrapped in the `layout_wrapper` and then you can get rid of the `max-breakpoint` stuff. (in the designs looks like everything should be the same width)
> 
> looks like there needs to be some margin at the very bottom as well. And looking at the design a bit more space between the primary callouts and the rest of the footer.

Got it! I'm not sure why I was thinking that the `Primary Callouts` needed to be a little wider than the rest of the footer. Maybe it use to be like that? Regardless, I just looked at the designs again, and I'll take care of it 👍

andylushman:
> This still looks a bit off from the design. I think the entire footer can be wrapped in the `layout_wrapper` and then you can get rid of the `max-breakpoint` stuff. (in the designs looks like everything should be the same width)
> 
> looks like there needs to be some margin at the very bottom as well. And looking at the design a bit more space between the primary callouts and the rest of the footer.

Should looks like the new designs now.

PR #593, opened 2019-07-16
Start to port venues form to new styling

jayzes:
- [ ] Venue new/edit form
- [ ] Venue availability form


app/views/components/_venue_card.html.slim
```diff
@@ -0,0 +1,11 @@
...
+  .VenueCard
+    .VenueCard-link
+      = link_to edit_venue_path(venue)  do
+        = render partial: 'icons/pencil'
+    h2.VenueCard-title #{venue.name}
+    .VenueCard-address
+      . #{venue.address}
+      . #{venue.city}, #{venue.state}
+    p.VenueCard-description #{venue.description}
+    .ContentCard-action
```
scpaye506: is this `.ContentCard-action` being used?

app/assets/stylesheets/redesign/components/venue_edit.sass
```diff
@@ -0,0 +1,41 @@
...
+
+.VenueEdit-availabilities
+  margin-top: 25px
+
+.VenueEdit-submit
+  +white-button(315px, 56px)
+  margin: 40px auto 0
+  width: 315px
+
+table
```
scpaye506: let's add classes to these table elements or nest them within some sort of class so these styles don't leak into global styles. Incase there are ever tables somewhere else

app/views/venues/edit.html.slim
```diff
@@ -1,19 +1,16 @@
-= render partial: 'dashboards/nav', locals: { item: :venue }
+- content_for(:fullscreen_takeover, true)
```
scpaye506: it's probably worth putting all of this into the venue form component and then being able to reuse the entire form in both the new and edit files (even if we aren't using new for now). You can follow the pattern setup in articles/articles_form and articles/new and articles/edit files

app/controllers/venues_controller.rb
```diff
@@ -4,11 +4,6 @@ class VenuesController < ApplicationController
   before_action :set_venue, only: %i[show edit update destroy]
```
scpaye506: add `skip_before_action :store_location, only: %i[new edit]` above this `before_action` forgot to mention this, but it helps with closing the fullscreen takeover screens

app/views/components/_venue_card.html.slim
andylushman: just removed it.

app/controllers/venues_controller.rb
andylushman: Sounds good. Done.

app/assets/stylesheets/redesign/components/venue_edit.sass
andylushman: Good call! Just created `VenueEdit-table` and nested the rest over it.

PR #600, opened 2019-07-16
Make articles/sessions go back to the page you came from

scpaye506:


scpaye506:
> This looks good except when we go from `Home` to an article and then click on another article from `YMABII` the back button will then alternate between the two articles instead of taking us back to the home page.

good point, that does complicate things a bit

PR #631, opened 2019-08-04
Users can now specify ethnicities at registration

### LINKED ISSUE #626: Add ethnicity question to registration form
Per conversation with @ryefinegan and Olivia. Using the US Census questions as a basis:

image

jayzes:
- From a dynamic list managed in the admin
- Implemented the same way as attendee goals
- Alphabetize factories.rb while we’re at it

Fixes #626

### DIFF
--- a/app/admin/ethnicities.rb
+++ b/app/admin/ethnicities.rb
@@ -1,0 +1,13 @@
+ActiveAdmin.register Ethnicity do
+  menu parent: "Site Content"
+
+  permit_params :name,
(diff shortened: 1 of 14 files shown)


app/admin/ethnicities.rb
```diff
@@ -0,0 +1,13 @@
+ActiveAdmin.register Ethnicity do
```
marlabrizel: nice call on putting this in the admin 👍

PR #669, opened 2019-09-08
Move recaptcha filter to article create action

jayzes:
Should catch slightly fewer folks this way while still stopping spam.

Also add v2 checkbox fallback.

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -26,12 +26,22 @@ class ArticlesController < ApplicationController
 
   def create
     @article = current_user.articles.new(article_params.merge(author_ids: [article_params[:author_ids]]))
-    if @article.save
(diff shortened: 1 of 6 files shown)


app/controllers/articles_controller.rb
```diff
@@ -26,12 +26,22 @@ def new
 
   def create
     @article = current_user.articles.new(article_params.merge(author_ids: [article_params[:author_ids]]))
-    if @article.save
-      flash[:notice] = "Thanks! Your article has been received."
-      NotificationsMailer.notify_of_new_article(@article).deliver_now
-      redirect_to dashboard_path
+    success = verify_recaptcha(action: "create_article", model: @article)
```
marlabrizel: can we call `success` something slightly more descriptive so that it's a little more obvious there are two layers of recaptcha passing?

PR #670, opened 2019-09-09
Add no index option for older sessions

jayzes:


jayzes:
@marlabrizel this should help on job fair

PR #674, opened 2019-09-11
Fix travis

jayzes:


jayzes:
@marlabrizel going to merge this - bumping the ruby version to the latest 2.5 fixes it. It does require installing a new version locally and doing that FFI fix again, though.

marlabrizel:
sigh. go for it

PR #683, opened 2019-10-25
Upgrade Rails to 6.0

marlabrizel:
Upgrade Rails to `6.0.0`.

This does _not_ upgrade Ruby to `latest` since `5.3.x` is already compatible with Rails 6. Not saying we shouldn't upgrade that too, just saying we should minimize complexity and upgrade that separately :)

I have tested this locally, but it merits another test-through on another environment to fully ensure there aren't regressions.

Specifically, I am not sure if I fully rectified issues related to sprockets changes (through creating the `manifest.js`) file. [This article](https://www.schneems.com/2017/11/22/self-hosted-config-introducing-the-sprockets-manifestjs/) suggests more changes could be needed but I couldn't trigger a use case locally.

Otherwise, there's not much to this - hooray for a robust test suite.


jayzes:
FYI @marlabrizel - https://github.com/denverstartupweek/dsw-site/pull/683/commits/02e4fb6c6b137e93d04fc1c89269234bbc4c7c74 is the reason that staging wasn't booting

PR #745, opened 2020-02-21
Revert "Bump sendgrid-ruby from 6.0.4 to 6.1.0"

jayzes:
Reverts denverstartupweek/dsw-site#742


jayzes:
Reverting because I think this may be causing https://app.honeybadger.io/projects/48239/faults/61056039

Issue #776, opened 2020-04-04
As someone with resources to give, I should be able to submit resources that I want to publicize


jayzes:
Modeled on our existing CFP/article flows, this should be a form that goes someplace in the admin for review and dissemination.

https://www.denverstartupweek.org/articles/new

Need to determine what information is collected. Definitely a user account needed. To start:
- Name
- Image
- Company Name
- Description
- Contact information (optional)
- Website
- Timeframe / expiration date

Should probably kick out a notification e-mail to info@denverstartupweek.org too.

jmkoni:
I have to roll off this project, so reassign if needed!

PR #807, opened 2020-05-01
Add support areas and industry types

jmkoni:
Part of #776

I know we talked about calling this "Areas of Support" but due to Rails and it's pluralizing, I decided it would be simpler to go with SupportArea. We can make the text the user sees "Areas of Support".

❓ How do we want to be able to create these? By console? Rake task? Is this through ActiveAdmin?

### DIFF
--- a/app/admin/industry_types.rb
+++ b/app/admin/industry_types.rb
@@ -1,0 +1,19 @@
+ActiveAdmin.register IndustryType do
+  permit_params :name
+
+  index do
+    selectable_column
+    column :name
+    actions
+  end
+
+  filter :name
+
+  form do |f|
+    f.inputs do
+      f.input :name
+    end
+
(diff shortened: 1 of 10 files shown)


app/admin/industry_types.rb
```diff
@@ -0,0 +1,19 @@
+ActiveAdmin.register IndustryType do
```
jayzes, line 1: We might want to nest these under another top-level category in the admin (otherwise that top nav gets super long). Here's an example: https://github.com/denverstartupweek/dsw-site/blob/master/app/admin/clusters.rb#L2. I think there's enough to this whole project that it might be worth creating a top-level `GiveToo` menu?

jmkoni: Great idea! Done :)

PR #898, opened 2020-08-27
Update scheduled_session.sass so Scheduled Session Title does not cut off hanging letters

s0rta:
Saw a design flaw on the schedule page, the .ScheduledSession-title class had a line height set to 1, this was causing g's to be cut off. Updated the class to have a line height of 1.1, negating the problem in the simplest way I could. 

I might not be comprehending the pull request doc but I think I followed all the rules. 

- JT


marlabrizel:
Thank you @s0rta - really appreciate it!

PR #912, opened 2020-09-01
Update information on the about page for 2019 stats

Jessiewithani:
Updated the about me page from 2018 stats to 2019 status. Also did some minor grammar fixes.


jayzes:
LGTM! We might want to remove the `.DS_Store` files (they're a weird OSX artifact that doesn't really need to be in the repo) but otherwise looks good to merge/deploy!

Jessiewithani:
it did that automatically, but I'll keep that in mind from now on. Alright
let me remove that file and then I'll let you know!

Issue #903, opened 2020-08-28
Homepage should display the new "glitched-out" 2020 logo & header


jayzes:
Invision mocks: https://projects.invisionapp.com/d/main#/projects/20492622

<img width="1418" alt="Screen Shot 2020-09-02 at 2 59 25 PM" src="https://user-images.githubusercontent.com/22783/92036059-ea17e580-ed2c-11ea-80c6-59055f779bf3.png">

PR #922, opened 2020-09-03
Styling

Jessiewithani:
these are the styling I've done so far. I've specified them by commits, but if you'd rather me separate them from branches I can redo it, no big deal. Or I can start to make separate branches after this. Let me know.

### DIFF
--- a/app/assets/stylesheets/redesign/core/base.sass
+++ b/app/assets/stylesheets/redesign/core/base.sass
@@ -40,6 +40,7 @@ main
   min-height: 100%
   margin: 0 auto
   padding-top: 111px
+  background-color: black
 
 main.is-fullscreen
   padding-top: 0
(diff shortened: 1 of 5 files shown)


jayzes:
Closes https://github.com/denverstartupweek/dsw-site/issues/903

jayzes:
@Jessiewithani just updated the base branch to merge into https://github.com/denverstartupweek/dsw-site/pull/923

jayzes:
Not sure how you want to handle review from a QA perspective? I'm happy to run through it here and we can figure out how to do a punchlist or something like that We should probably eventually run things past Kelsey too, I feel like it's always good to get a designer's eye on things too.

The big one that jumps out at me is that I think we'll want to adjust the black background so it has the angled lower edge that transitions to the white background for the lower sections - basically this bit in the InVision mocks:

Screen Shot 2020-09-03 at 1 04 59 PM

As I'm looking at it locally, I'm also realizing that we'll need to do something with the block of homepage CTAs:

Screen Shot 2020-09-03 at 1 14 48 PM

I'll ping Kelsey about that one in Slack right now.

That aside, this is looking like a great start for sure!

PR #954, opened 2020-09-09
Stub out live/upcoming session homepage cards

jayzes:
Add a `live_and_upcoming` scope to Submissions

Still needs a few more tests to ensure that upcoming gets pulled properly but it's a start

### DIFF
--- a/app/views/layouts/shared/_live_streaming.html.slim
+++ b/app/views/layouts/shared/_live_streaming.html.slim
@@ -1,1 +1,5 @@
 div.Live_stream_video_container
+  - Submission.live_and_upcoming.livestreamed.each do |session|
+    = render 'components/three_card_list' do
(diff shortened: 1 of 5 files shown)


app/views/layouts/shared/_live_streaming.html.slim
```diff
@@ -1 +1,5 @@
-div.Live_stream_video_container
\ No newline at end of file
+div.Live_stream_video_container
+  - Submission.live_and_upcoming.livestreamed.each do |session|
+    = render 'components/three_card_list' do
+      = render 'components/three_card_list_item' do
+        = render 'components/session_card', session: session
```
jayzes, line 5: @Jessiewithani we'll probably need to create a new kind of card from here, but I figured this would give us a place to start.

PR #969, opened 2020-09-11
Allow users to override the homepage live time

jayzes:
You can now pass in an ISO8601-formatted `as_of` date time parameter and that’ll get passed all the way through.

e.g. http://localhost:3000/?as_of=2020-09-11T15:18:51-06:00


Jessiewithani:
This will make things easier to test out now. Thanks Jay.

PR #972, opened 2020-09-12
Jessle/homepage complete

Jessiewithani:
I set the components with different background colors for now so it's easier to see what you're dealing with. Let me know if you'd rather work with it in one color.


Jessiewithani:
I fixed the merge conflict and merged it just like we agreed. Let me know if things aren't working on your end.

PR #982, opened 2020-09-14
Fix margin for session cards

### LINKED ISSUE #973: SessionDetail-presenter could use a bit of padding to match SessionDetail-attendee
Minor nit, but it'd be good to even out. Presenters are on the left, attendees on the right (this is a session detail page like https://www.denverstartupweek.org/schedule/6923-from-hello-world-to-vp-of-engineering-with-nick-caldwell-twitter-moderated-by-aaron-clark-justice-reskill)

Screen Shot 2020-09-11 at 9 30 28 PM

Jessiewithani:
Added margin instead of padding to the name cards. Looks very neat. Let me know if this is okay with you. This should close #973

### DIFF
--- a/app/assets/stylesheets/redesign/components/name_with_avatar.sass
+++ b/app/assets/stylesheets/redesign/components/name_with_avatar.sass
@@ -3,6 +3,7 @@
 .NameWithAvatar
   display: flex
   align-items: center
+  margin-top: 10px
 
 .NameWithAvatar-name
   @include font-family-mixin(Montserrat-Light)


jayzes:
Just did a bit of poking with this style on production (e.g. https://www.denverstartupweek.org/schedule/6924-denver-startup-week-kickoff-celebration-and-keynote-with-robert-smith-sponsored-by-wix) - I think this doubles up the padding on   `SessionDetail-attendee` unfortunately. What if we added a `margin: 6px 0` to `SessionDetail-presenter` instead? I think that'd make the presenter side match the attendee side without spacing out the attendees even further.

jayzes:
Going to close this as I think it's supeceded by #986

PR #1012, opened 2020-09-19
Fix schdule list container

### LINKED ISSUE #999: Schedule list items are occasionally breaking the layout
See the second item here (from the schedule page):

Screen Shot 2020-09-17 at 11 23 30 AM

Jessiewithani:
The schedule container shouldn't break anymore, and I tried it out with the media queries and it looks fine as well. I added an overflow scroll just in case theres a title that is ridiculously long and added a higher max-width as well.  This should close #999


jayzes (approved):
Looks great!

PR #1381, opened 2022-05-28
tawheels/remove vote count 1380

tawheels:
Added logic to only allow user to vote if they have not voted
grey out the thumb when user has voted for the session


app/models/submission.rb
```diff
@@ -409,6 +409,14 @@ def company_name=(value)
     self.company = Company.where("LOWER(name) = LOWER(?)", value).first_or_initialize(name: value)
   end
 
+  def already_voted(current_user)
+    if current_user
+      current_user.votes.exists?(:submission_id => id)
```
jayzes, line 414: This is a total nit, but we've generally moved to the newer-style colon syntax for hash literals in the rest of the codebase, might make sense to do the same here.

```suggestion
      current_user.votes.exists?(submission_id: id) 
```

jayzes (approved):
One nit, otherwise looks great!

Issue #1398, opened 2022-07-13
Add an optional link to open positions to the Job Fair signup?


tawheels:
also "attending slots" can be deleted or hidden, it is not relevant to the in-person version of the job fair

PR #1400, opened 2022-07-13
Add an optional link to open positions to the Job Fair signup?

tawheels:
also "attending slots" can be deleted or hidden, it is not relevant to
the in-person version of the job fair

Issue #1398


tawheels:
10-4 on all, I will remove the slots, and you figured out that AWS was defined below in the credentials. When tests pass I will push to heroku.

